PR #30, opened 2022-04-13
Optimobile foundations

BBHOptimove:
Add foundations of Optimobile functionality by combining base elements from the Kumulos SDK into the Optimove SDK.


BBHOptimove:
optimobile pr to dev

Optimobile/AppDelegate.swift
```diff
@@ -0,0 +1,62 @@
...
+//  Created by Barak Ben Hur on 04/04/2022.
+//
+
+import UIKit
+import KumulosSDK
+
+@main
+class AppDelegate: UIResponder, UIApplicationDelegate {
+
+    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
```
optimove-andy: Im not sure im following yet - this file looks like the AppDelegate for an app and has storyboards etc? Should this all be deleted or is it a test app shipping with the SDK? From the diff it seems to be new.

OptimoveSDK/Sources/Classes/Optimobile.swift
```diff
@@ -0,0 +1,129 @@
...
+    
+    public static var sheard: Optimobile {
+        get {
+            return internalQueue.sync {
+                _sheard
+            }
+        }
+    }
+    
+    public typealias deepLinkComplition = (_ deepLinkResolution: DeepLinkResolution) -> ()
```
optimove-andy: minor spelling note "Complition" -> "Completion"

OptimoveSDK/Sources/Classes/Optimobile.swift
```diff
@@ -0,0 +1,129 @@
...
+             pushOpened(responder: pushOpenedComplition?)
+    }
+    
+    private static var builder: KSConfigBuilder!
+    
+    private init() {}
+    
+    public static func configure(for tenantInfo: OptimoveTenantInfo, apiKey: String, secretKey: String, abilities: [Abilities]? = nil) {
+        Optimove.configure(for: tenantInfo)
+        builder = KSConfigBuilder(apiKey: apiKey, secretKey: secretKey)
```
optimove-andy: Maybe loop Kosta in on this question but i think we should look to hide any Kumulos / KS references, even if we just renamed the class to something like OptiMobile / OptimobileConfigBuilder

Optimobile/AppDelegate.swift
cgwyllie: I think these files were used during prototyping, have removed them form the PR.

OptimoveSDK/Sources/Classes/Optimobile.swift
cgwyllie: Renaming passes to follow

OptimoveSDK/Sources/Classes/Observers/AppOpenObserver.swift
```diff
@@ -1,60 +0,0 @@
-//  Copyright © 2019 Optimove. All rights reserved.
-
-import Foundation
-import OptimoveCore
-import UIKit.UIApplication
-
-final class AppOpenObserver {
```
cgwyllie: I'm not sure if this behaviour should be removed, is it Optipush-specific?

cgwyllie: Restored

OptimoveSDK/Sources/Classes/Optimobile.swift
cgwyllie: Renaming pass to follow when combining the public APIs

PR #34, opened 2022-04-21
Restructure notification service extension target & shared code

cgwyllie:
The Shared symlink was causing some issues and had duplicated files in a few places.

The OptimoveNotificationServiceExtension target still had Optipush code in it as well as the Kumulos SDK extension code.

Remove and add a new target for the extension that only contains the Kumulos SDK sources.

Renaming to follow.

Note this likely breaks packaging for pods etc., expect to pick up packaging in a follow-up PR.


Optimove.xcodeproj/xcshareddata/xcschemes/OptimoveCore.xcscheme
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cgwyllie, line 1: I'm not 100% these schemes are required, although for Carthage certainly I think we need shared schemes.

There were some schemes removed in the `Optimobile` branch (need to check which), but I was just auto creating schemes in Xcode and these have come back.

cgwyllie: OK yes, these schemes were previously present in the dev branch, and have been re-created now. Figure we probably want them, once merged to Optimobile we'll see any diff with `dev`.

PR #35, opened 2022-04-22
Refactor SDK config to support both Optimove and Optimobile configurations

cgwyllie:
Promote and extend KSConfigBuilder as OptimoveConfigBuilder. It now expects to be given credentials in the base64-encoded JSON tuple provided in the mobile UI.

Refactor Kumulos SDK classes to expect an OptimobileConfig, held in the OptimoveConfig.

Add a new initialize method that conditionally inits the configured SDK behaviours.

Note there is follow-up work to merge the public API of Optimobile into the Optimove class, so it is commented out in this build.


OptimoveSDK/Sources/Classes/Optimobile.swift
```diff
@@ -11,94 +11,94 @@ import CoreLocation
 
 public class Optimobile {
     
-    private static let internalQueue = DispatchQueue(label: "com.singletioninternal.queue",
```
cgwyllie, line 14: This is just commented for now to get the build passing, it will be flattened out in follow-up work.

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -0,0 +1,203 @@
...
+    @objc public var tenantToken: String
+    @objc public var configName: String
+
+    @objc public init(tenantToken: String, configName: String) {
+        self.tenantToken = tenantToken
+        self.configName = configName
+    }
+}
+
+public struct OptimobileConfig {
```
cgwyllie, line 35: For Objective-C interop we'll need to refactor these as NSObject classes. There's a separate task for the same.

PR #37, opened 2022-04-22
Guard setUserId calls based on SDK configuration

cgwyllie:
setUserId and registerUser should report to both Optimove and Optimobile backend systems if both are configured.

If only one or the other is configured, reporting should be limited to the configured backend system.

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimove.swift
+++ b/OptimoveSDK/Sources/Classes/Optimove.swift
@@ -22,6 +22,7 @@ typealias Logger = OptimoveCore.Logger
     }()
 
     private let container: Container
+    private var config: OptimoveConfig!
 
     private override init() {
         self.container = Assembly().makeContainer()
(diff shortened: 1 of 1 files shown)


OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -22,6 +22,7 @@ typealias Logger = OptimoveCore.Logger
     }()
 
     private let container: Container
+    private var config: OptimoveConfig!
```
cgwyllie, line 25: This maybe isn't the neatest way to do this, but I couldn't see how to easily provide any arguments to the shared initialiser block.

As usage is very constrained, perhaps it's acceptable.

optimove-andy: I can live with this, its easier than making nullable and checking every time we try to use it.

PR #38, opened 2022-04-22
Use Optimove initialVisitorId for Optimobile installId

cgwyllie:
Refactor the Kumulos installId to use the Optimove initialVisitorId instead.

The initialVisitorId is also a UUIDv4 so this change should be transparent to the Optimobile implementation.

In order to achieve this in cases where Optimove credentials are not yet provided, we need to re-arrange the initialisation of some dependencies in the container.

Move the Logger init from configure, and the new visitor ID generation from startSdk into the Optimove init() method instead.

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimove.swift
+++ b/OptimoveSDK/Sources/Classes/Optimove.swift
@@ -26,6 +26,10 @@ typealias Logger = OptimoveCore.Logger
 
     private override init() {
         self.container = Assembly().makeContainer()
+        container.resolve { serviceLocator in
+            serviceLocator.loggerInitializator().initialize()
+            serviceLocator.newVisitorIdGenerator().generate()
+        }
         super.init()
     }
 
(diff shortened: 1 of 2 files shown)


OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -26,6 +26,10 @@ typealias Logger = OptimoveCore.Logger
 
     private override init() {
         self.container = Assembly().makeContainer()
+        container.resolve { serviceLocator in
```
cgwyllie, line 29: Based on my current understanding of the container, this change should be OK.

Perhaps the logger init could stay inside configure(for:) but I wasn't 100% sure and the logger doesn't seem to be able to init more than once so I moved it. It means the logger should be useable inside the visitor ID generator if required.

Visitor ID generation is quite simple and is safe to be run multiple times so it could have remained in both places, but I didn't see a need to duplicate the work.

PR #45, opened 2022-04-26
Remove Optipush code

cgwyllie:
Remove Optipush code and related test cases.


OptimoveCore/Sources/Classes/Storage/StorageFacade.swift
```diff
@@ -23,14 +23,11 @@ public enum StorageKey: String, CaseIterable {
     case advertisingIdentifier
     case optFlag
```
cgwyllie: I'm not sure what this flag is for and if it was related to Optipush or not.

OptimoveSDK/Sources/Classes/Initializer/SDKInitializer.swift
```diff
@@ -60,9 +60,7 @@ private extension SDKInitializer {
 
         // 4 responder
         let componentHanlder = ComponentHandler(
```
cgwyllie, line 62: I'm not sure if this component handler is required for setup or not.

Barak's branch had this block commented out, it wasn't clear if this is _only_ related to Optipush (does it just decorate the other components/event builder?) or if it is still needed to ensure the other optistream components and event builders are correctly set up.

k-antipochkin: The Optipush component here should be relevant to Optipush only

cgwyllie:
Ok so is it fine to remove the whole componentHandler declaration?

OptimoveSDK/Sources/Classes/Initializer/SDKInitializer.swift
cgwyllie: Ok so is it fine to remove this componentHandler declaration entirely?

OptimoveCore/Sources/Classes/Storage/StorageFacade.swift
cgwyllie: Kostya confirmed related to Optipush, have since removed.

OptimoveSDK/Sources/Classes/Initializer/SDKInitializer.swift
cgwyllie: So think it shouldn't be removed entirely, reading through it looks like if the ComponentHandler was removed then the optistreamComponents and OptistreamEventBuilder wouldn't be wired into the pipeline at all.

PR #47, opened 2022-04-28
Tidy up MobileProvision implementation

cgwyllie:
The ported code from Objective-C has been factored out to use the existing Codable parser
implementation.

Mapping of simulator release type was removed seeing as simulators don't currently support
push, no tokens are given anyway.

This was done to ensure support for iOS 10+, not just iOS 13+.

References:

- https://developer.apple.com/documentation/bundleresources/entitlements/aps-environment
- https://developer.apple.com/forums/thread/685723
- https://stackoverflow.com/a/5764504

### DIFF
--- a/OptimoveSDK/Sources/Classes/Provisioning/MobileProvision.swift
+++ b/OptimoveSDK/Sources/Classes/Provisioning/MobileProvision.swift
@@ -4,11 +4,12 @@ import Foundation
 import OptimoveCore
 
 struct MobileProvision: Decodable {
-
-    var entitlements: Entitlements
+    let entitlements: Entitlements
+    let provisionsAllDevices: Bool?
 
     private enum CodingKeys: String, CodingKey {
         case entitlements = "Entitlements"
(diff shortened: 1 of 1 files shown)


OptimoveSDK/Sources/Classes/Provisioning/MobileProvision.swift
```diff
@@ -77,74 +78,20 @@ extension MobileProvision {
...
-    
     static func releaseMode() -> UIApplicationReleaseMode {
-        var entitlements: NSDictionary
-        let mobileProvision = getMobileProvision()
-        
-        guard mobileProvision != nil else { return .unknown }
-        
-        entitlements = mobileProvision!.object(forKey: "Entitlements") as! NSDictionary
-        
-        if mobileProvision!.count == 0 {
```
cgwyllie, line 132: Not 100% sure if this is equivalent, although I suspect the 'empty dict' is now covered by having a nil MobileProvision struct because the parsing of Entitlements fails.

PR #49, opened 2022-04-28
In-app sync behaviour changes

cgwyllie:
- [Don't block caller when performing in-app syncs](https://github.com/optimove-tech/Optimove-SDK-iOS/pull/49/commits/8f520c5fc21b777fe06eed9a9c78800a6c15b871)
- [Don't perform unnecessary in-app syncs from content-available pushes](https://github.com/optimove-tech/Optimove-SDK-iOS/pull/49/commits/15ae8a7616b0187ccd125d508843584a9ec84861)

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
+++ b/OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
@@ -404,55 +404,61 @@ class PushHelper {
         let kumulosDidFailToRegister = imp_implementationWithBlock(didFailToRegBlock as Any)
         existingDidFailToReg = class_replaceMethod(klass, didFailToRegisterSelector, kumulosDidFailToRegister, didFailToRegType)
 
-        // iOS9 did receive remote delegate
         // iOS9+ content-available handler
         let didReceiveSelector = #selector(UIApplicationDelegate.application(_:didReceiveRemoteNotification:fetchCompletionHandler:))
         let receiveType = NSString(string: "v@:@@@?").utf8String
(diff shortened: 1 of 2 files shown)


OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
```diff
@@ -404,55 +404,61 @@ class PushHelper {
...
+            let hasInApp = notification.inAppDeepLink() != nil
+
+            self.setBadge(userInfo: userInfo)
+            self.trackPushDelivery(notification: notification)
+
+            if existingDidReceive == nil && !hasInApp {
+                // Nothing to do
+                completionHandler(.noData)
+                return
+            } else if existingDidReceive != nil && !hasInApp {
```
cgwyllie, line 421: We could chop this escape hatch but seemed less overhead to have it at the cost of more verbosity.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
```diff
@@ -404,55 +404,61 @@ class PushHelper {
...
+
+            var fetchResult : UIBackgroundFetchResult = .noData
+            let group = DispatchGroup()
+
+            if existingDidReceive != nil {
+                group.enter()
+                DispatchQueue.main.async {
+                    unsafeBitCast(existingDidReceive, to: kumulos_applicationDidReceiveRemoteNotificationFetchCompletionHandler.self)(obj, didReceiveSelector, application, userInfo, { (result : UIBackgroundFetchResult) in
+                        DispatchQueue.main.async {
+                            if fetchResult == .noData {
```
cgwyllie, line 435: Result handling is also sent back to the main thread to ensure thread-safety mutating the var (we don't know what thread the user may invoke the callback from).

Potentially could test if we're already main to reduce redundant work.

PR #53, opened 2022-05-02
Objective-C support for Optimobile features

BBHOptimove:
-  OptimoveConfigBuidler & OptimoveConfig annotated with @objc and derive from NSObject
- New Optimove and OptimoveInApp APIs annotated with @objc and derive from NSObject
- Deep link handler uses associated value enums, expose `enableDeepLinking` overload on config builder that turns value into a dictionary for ObjC handler


OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -121,6 +155,14 @@ open class OptimoveConfigBuilder: NSObject {
 
         return self
     }
+    
+    
+    @objc @discardableResult public func enableDeepLinkingObjC(cname: String? = nil, _ objCHandler: @escaping (NSDictionary) -> Void) -> OptimoveConfigBuilder {
+         _deepLinkCname = URL(string: cname ?? "")
+         _objCdeepLinkHandler = objCHandler
```
cgwyllie: Hi @BBHOptimove, I'm not sure this is a complete solution because nothing currently calls this handler from inside the SDK with the deep link resolution. The current implementation in Optimove+DeepLinking.swift will call the OptimobileConfig.deepLinkHandler with a resolution when a deferred deep link is triggered. E.g.

- User visits example.lnk.click/promotion
- Universal Link triggers delegate
- Delegate calls SDK handler
- SDK looks up promotion details
- SDK calls deep link handler block with deep link resolution

I was thinking something more along these lines might work to bridge the implementation:

```swift 
@objc @discardableResult public func enableDeepLinking(cname: String? = nil, _ objCHandler: @escaping (NSDictionary) -> Void) {
    return self.enableDeepLinking(cname: cname, { resolution in
        objcHandler(resolution.asDict())
    }
}
```

We can implement some kind of internal helper on deep link resolution like `asDict()->NSDictionary` that performs the mapping to the dictionary. Then from the SDK's point of view, the config only ever holds one callback reference (the existing `_deepLinkHandler`) and the change is transparent to the SDK.

Happy to discuss further if there's something I'm missing.

OptimoveSDK/Sources/Classes/Optimobile/OptimoveInApp.swift
```diff
@@ -3,7 +3,7 @@
 import Foundation
 import CoreData
 
-public class InAppInboxItem {
+@objc public class InAppInboxItem :NSObject {
```
cgwyllie, line 6: This looks good but are we missing any `@objc` annotations / `NSObject` subclassing on the public APIs in here (`OptimoveInApp`) (and the ones added to the Optimove class itself?)

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
BBHOptimove: @objc @discardableResult public func enableDeepLinking(cname: String? = nil, _ objCHandler: @escaping (NSDictionary) -> Void) {
    return self.enableDeepLinking(cname: cname, { resolution in
        objcHandler(resolution.asDict())
    }
}

i dont understand where to write that code

OptimoveSDK/Sources/Classes/Optimobile/OptimoveInApp.swift
BBHOptimove: added annotations to OptimoveInApp
Optimove class looks fine annotations wise

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
BBHOptimove: fixed

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -151,9 +192,10 @@ open class OptimoveConfigBuilder: NSObject {
                 inAppConsentStrategy: _inAppConsentStrategy,
                 inAppDeepLinkHandlerBlock: _inAppDeepLinkHandlerBlock,
                 pushOpenedHandlerBlock: _pushOpenedHandlerBlock,
-                _pushReceivedInForegroundHandlerBlock: _pushReceivedInForegroundHandlerBlock,
+                pushReceivedInForegroundHandlerBlock: _pushReceivedInForegroundHandlerBlock as Any,
```
cgwyllie: Why did we add the cast here, is it required?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -121,6 +155,13 @@ open class OptimoveConfigBuilder: NSObject {
 
         return self
     }
+    
+    @objc @discardableResult public func enableDeepLinking(cname: String? = nil, _ objcHandler: @escaping (NSDictionary) -> Void) -> OptimoveConfigBuilder {
+        _objCdeepLinkHandler = objcHandler
```
cgwyllie, line 160: Is it necessary to store the objc handler as a field, can we just capture it in the closure scope and save adding a field for it?

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+DeepLinking.swift
```diff
@@ -32,6 +32,20 @@ public enum DeepLinkResolution {
...
+    
+    func toDict() -> NSDictionary {
+        let dict = NSDictionary()
+        
+        switch self {
+        case let .lookupFailed(url): dict.setValue(url, forKey: "lookupFailed")
+        case let .linkNotFound(url): dict.setValue(url, forKey: "linkNotFound")
+        case let .linkExpired(url): dict.setValue(url, forKey: "linkExpired")
+        case let .linkLimitExceeded(url): dict.setValue(url, forKey: "linkLimitExceeded")
+        case let .linkMatched(data): dict.setValue(data, forKey: "linkMatched")
```
cgwyllie, line 44: Do we need to go deeper into converting 'data' to a dict as well, I noticed it was a DeepLink struct

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
cgwyllie: Great, LGTM!

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
BBHOptimove: i think it is better for memory handling purposes

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+DeepLinking.swift
BBHOptimove: i have looked at the deeper class׳s and structs and it looks like they confirm to the objc needs

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
BBHOptimove: _pushReceivedInForegroundHandlerBlock was not annotated as optional
Fixed

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
cgwyllie: Oh cool, could you explain why / share a reference so I can learn more about this?

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+DeepLinking.swift
cgwyllie: OK, you mean we can expose this struct to Objective-C?

I possibly misunderstood but I thought Objective-C can't use struct types from Swift at all--am I mistaken?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -151,9 +192,10 @@ open class OptimoveConfigBuilder: NSObject {
                 inAppConsentStrategy: _inAppConsentStrategy,
                 inAppDeepLinkHandlerBlock: _inAppDeepLinkHandlerBlock,
                 pushOpenedHandlerBlock: _pushOpenedHandlerBlock,
-                _pushReceivedInForegroundHandlerBlock: _pushReceivedInForegroundHandlerBlock,
+                pushReceivedInForegroundHandlerBlock: _pushReceivedInForegroundHandlerBlock,
                 deepLinkCname: _deepLinkCname,
                 deepLinkHandler: _deepLinkHandler,
+                objCdeepLinkHandler: _objCdeepLinkHandler,
```
cgwyllie, line 198: I'm not sure we need to pass this down into OptimobileConfig, it looks like it just needs held at the OptimoveConfigBuilder level? (the closure is passed from _deepLinkHandler)

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
barakbenhur1: because swift arc (automatic count reference) will keep hold of the class (it will be strong referenced) and it could lead to retain cycle

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
barakbenhur1: this code seems to not exist any more

PR #61, opened 2022-07-04
Notify Optimobile about existing user associations

optimove-andy:
### Description of Changes

During initialization the Optimobile backend should be notified of any existing pre-Optimobile user associations.

### Breaking Changes

-   None

### Test Plan

1. Init with Optimove credentials only, setUserId
2. Init with OptiMobile credentials, observe current userId being associated during initializations

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimove.swift
+++ b/OptimoveSDK/Sources/Classes/Optimove.swift
@@ -58,9 +58,7 @@ typealias Logger = OptimoveCore.Logger
                     return
                 }
                 
-                guard let userId = try? serviceLocator.storage().getCustomerID() else {
-                    userId = nil
-                }
+                let userId = try? serviceLocator.storage().getCustomerID()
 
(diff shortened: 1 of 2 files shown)


OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -57,8 +57,12 @@ typealias Logger = OptimoveCore.Logger
                 guard let visitorId = try? serviceLocator.storage().getInitialVisitorId() else {
                     return
                 }
+                
+                guard let userId = try? serviceLocator.storage().getCustomerID() else {
+                    userId = nil
+                }
```
cgwyllie: I don't think we need to guard this and re-assign, should an assignment to the `try?` statement not just be `String?`?

optimove-andy: 👍 updated

k-antipochkin (approved):

PR #72, opened 2023-08-28
Add in-app display mode to enable pausing & resuming in-app message display

cgwyllie:
### Description of Changes

Add an in-app display mode configuration that allows pausing display of in-app messages.

All existing in-app behaviours such as queueing are performed as normal whilst display is paused, only the presentation of the message view itself is prevented.

This is exposed as a 'default' configuration option and a new API to set the behaviour at runtime.

When display was previously paused and then enabled again, any pending messages in the queue should be presented.

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [x] `OptimoveCore.podspec`
- [x] `OptimoveSDK.podspec`
- [ ] ~`README.md`~
- [x] `CHANGELOG.md`
- [ ] ~Update major version numbers in wiki (basic integration + push guides)~

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [x] Register for push
- [x] Opt-in for In-App
- [x] Send test push
- [x] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

- [x] Squash and merge to master
- [x] Delete branch once merged

### Post Release:

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
+++ b/OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
@@ -28,14 +28,39 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
     private var messageQueue : NSMutableOrderedSet
     private var pendingTickleIds : NSMutableOrderedSet
 
+    private var displayMode : InAppDisplayMode
     private var currentMessage : InAppMessage?
 
-    override init() {
+    init(displayMode: InAppDisplayMode) {
         self.messageQueue = NSMutableOrderedSet.init(capacity: 5)
         self.pendingTickleIds = NSMutableOrderedSet.init(capacity: 2)
         self.currentMessage = nil
+        
+        self.displayMode = displayMode
 
         super.init()
+    }
+    
+    func setDisplayMode(_ mode: InAppDisplayMode) {
+        runOnMainThreadSync {
+            let resumed = mode != displayMode && mode != .paused
+
+            displayMode = mode;
+
+            if resumed {
+                presentFromQueue()
+            }
+        }
+    }
+
+    func getDisplayMode() -> InAppDisplayMode {
+        var mode : InAppDisplayMode = .automatic
+
+        runOnMainThreadSync {
+            mode = displayMode
+        }
+
+        return mode
     }
 
     func queueMessagesForPresentation(messages: [InAppMessage], tickleIds: NSOrderedSet) {
@@ -88,29 +113,44 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
             }
         }
 
+        let notShowingCurrentTickle = self.currentMessage != nil
+            && self.currentMessage!.id != (self.messageQueue[0] as! InAppMessage).id
+            && (self.messageQueue[0] as! InAppMessage).id == self.pendingTickleIds[0] as! Int64
+
+        let queueNotEmptyAndNotShowingAnything = self.currentMessage == nil && self.messageQueue.count > 0
+
+        let shouldShowSomething = notShowingCurrentTickle || queueNotEmptyAndNotShowingAnything
+
         messageQueueLock.signal()
-      
-        DispatchQueue.main.async {
-            self.initViews()
-            
-            if (self.currentMessage != nil
-                && self.currentMessage!.id != (self.messageQueue[0] as! InAppMessage).id
-                && (self.messageQueue[0] as! InAppMessage).id == self.pendingTickleIds[0] as! Int64) {
+
+        if (shouldShowSomething) {
+            DispatchQueue.main.async {
                 self.presentFromQueue()
             }
         }
     }
     
     func presentFromQueue() -> Void {
-        if (self.messageQueue.count == 0) {
+        messageQueueLock.wait()
+        defer {
+            messageQueueLock.signal()
+        }
+
+        if (self.messageQueue.count == 0 || displayMode == .paused) {
+            DispatchQueue.main.async {
+                self.destroyViews()
+            }
+
             return;
         }
         
-        if let loadingSpinner = self.loadingSpinner {
-            loadingSpinner.performSelector(onMainThread: #selector(UIActivityIndicatorView.startAnimating), with: nil, waitUntilDone: true)
+        self.currentMessage = (self.messageQueue[0] as! InAppMessage);
+
+        runOnMainThreadSync {
+            initViews()
+            self.loadingSpinner?.startAnimating()
         }
 
-        self.currentMessage = (self.messageQueue[0] as! InAppMessage)
         let content = NSMutableDictionary(dictionary: self.currentMessage!.content)
         content["region"] = Optimobile.sharedInstance.config.region
 
@@ -130,9 +170,6 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
         }
         
         messageQueueLock.wait()
-        defer {
-            messageQueueLock.signal()
-        }
         
         messageQueue.removeObject(at: 0)
         pendingTickleIds.remove(message.id)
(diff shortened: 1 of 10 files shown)


OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -130,38 +160,31 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
         messageQueueLock.wait()
-        defer {
-            messageQueueLock.signal()
-        }
              
         self.messageQueue.removeAllObjects()
         self.pendingTickleIds.removeAllObjects()
         self.currentMessage = nil
+
+        messageQueueLock.signal()
```
cgwyllie, line 202: It didn't seem necessary to hold this lock over the view cleanup

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -318,29 +341,23 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
         let type = body["type"] as! String
         
         if (type == "READY") {
-              messageQueueLock.wait()
-              defer {
-                  messageQueueLock.signal()
-              }
```
cgwyllie, line 324: This lock is now acquired & released in presentFromQueue directly to save callers needing to reason about upcoming queue operations

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -130,38 +160,31 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
-            messageQueueLock.signal()
-        }
         
         messageQueue.removeObject(at: 0)
         pendingTickleIds.remove(message.id)
         currentMessage = nil
         
         if messageQueue.count == 0 {
             pendingTickleIds.removeAllObjects()
-            self.destroyViews()
```
cgwyllie, line 143: View cleanup is now included in presentFromQueue().

This allows simplifying the conditional nesting below, and it is more consistent with the Android SDK implementation so it is easier to compare the changes for equivalence.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -130,38 +160,31 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
-        }
-        else {
-            presentFromQueue()
         }
+
+        messageQueueLock.signal()
```
cgwyllie, line 190: The lock is now also acquired inside presentFromQueue() so needs to be released prior to entering that scope.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -88,29 +103,44 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
+            && (self.messageQueue[0] as! InAppMessage).id == self.pendingTickleIds[0] as! Int64
+
+        let queueNotEmptyAndNotShowingAnything = self.currentMessage == nil && self.messageQueue.count > 0
+
+        let shouldShowSomething = notShowingCurrentTickle || queueNotEmptyAndNotShowingAnything
+
         messageQueueLock.signal()
-      
-        DispatchQueue.main.async {
-            self.initViews()
```
cgwyllie, line 94: Previously we always initialised an in-app view, presuming that the queue will now hold something to show (it should).

The in-app view would then signal READY back to the host, at which point presentFromQueue was called.

This effectively tied initial queue presentation to the view state.

When adding the paused state, this meant the loading spinner would briefly flicker whilst the webview loads, but then presentFromQueue would destroy the views when the presentation is paused.

So it seems it's both unnecessary and undesirable to couple view init & initial presentation together.

By removing the call here, and only calling from presentFromQueue, it enables paused state to be respected in one place and removes the 'flickering'.

It does mean that we could post the DISPLAY_MESSAGE command to the webview before it is READY on the initial presentation, but practically it doesn't seem to matter because the READY event calls presentFromQueue again, and it's idempotent.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -88,29 +103,44 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
-            
-            if (self.currentMessage != nil
-                && self.currentMessage!.id != (self.messageQueue[0] as! InAppMessage).id
-                && (self.messageQueue[0] as! InAppMessage).id == self.pendingTickleIds[0] as! Int64) {
```
cgwyllie, line 98: This logic is concerned with the following edge-case:

- the app is showing an in-app message
- the user taps a tickle for a different in-app
- the user should see the tapped tickle immediately
- the SDK may still be loading the message referenced by the tickle (hence 'pending')

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -88,29 +103,44 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
+
+        if (shouldShowSomething) {
```
cgwyllie, line 127: We change the logic here to always call presentFromQueue if:

- there's a pending tickle that differs from the current message being displayed, or
- there's not currently a message being shown but we have a non-empty queue

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
cgwyllie: Confirmed, the JS evaluation on the first time load will result in the following error:

```
Optional(Error Domain=WKErrorDomain Code=4 "A JavaScript exception occurred" UserInfo={WKJavaScriptExceptionLineNumber=1, WKJavaScriptExceptionMessage=ReferenceError: Can't find variable: postHostMessage, WKJavaScriptExceptionColumnNumber=16, WKJavaScriptExceptionSourceURL=undefined, NSLocalizedDescription=A JavaScript exception occurred})
```

It doesn't seem to cause any fault, but in Android there was a refactor to hold a state machine over the view, perhaps we should introduce some 'ready state' tracking here as well?

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -28,15 +28,30 @@ class InAppPresenter : NSObject, WKScriptMessageHandler, WKNavigationDelegate{
...
         self.pendingTickleIds = NSMutableOrderedSet.init(capacity: 2)
         self.currentMessage = nil
+        
+        self.displayMode = displayMode
 
         super.init()
     }
+    
+    func setDisplayMode(_ mode: InAppDisplayMode) {
+        runOnMainThreadSync {
```
cgwyllie, line 46: Using this to serialise writes to the value.

Reads 'usually' appear from main thread, but not traced if they _always_ do yet.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
vvoicehovics: could we init views only if state is not paused? then also quit before showing spinner if it became paused on next present

`unnecessary and undesirable to couple view init & initial presentation together.` -- it's still coupled, but with + failing call before. And in any case it will remain coupled if we need to wait for 'ready'

But guess moving it to presentFromQueue makes sure views are initialised before any presentation. So, `setDisplayMode` doesnt need to care

except that js error look good

cgwyllie: Yes, we already won't init views (now) if presentation is paused, this is a slightly different condition that was exposed by the work.

In Android SDK you can see how we handle this in the InAppView, it maintains a state machine over 'loading' -> 'ready' -> 'disposed', and it will only pass messages to the renderer once ready is achieved.

Here we don't have the distinctions and the refactoring results in us naively trying to pass the message to the view.

We could potentially explore a similar state machine or 'ready' flag for the webview and skip passing messages if not ready, but I'm not sure if practically this error affects much.

WDYT?

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
cgwyllie: Reads are now done through the accessor on the main thread.

vvoicehovics (approved):

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
cgwyllie: Have added a simple ready state flag to track webview initialisation and prevent this error.

PR #78, opened 2023-11-19
feat(API): add track Open Notification metric API

eligutovsky:
### Description of Changes

Introducing a public API method that sends the open notification event by calling `Optimove.trackOpenMetric(userInfo: [AnyHashable:Any])`. Some apps have their own UserNotificationCenter delegate, which prevents our SDK from sending metrics.

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [ ] `OptimoveCore.podspec`
- [ ] `OptimoveSDK.podspec`

- [ ] `README.md`
- [ ] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -367,6 +367,20 @@ extension Optimove {
...
+    /// - Parameter notificationId: the notification ID
+    @objc public func trackOpenMetric(notificationId: Int) {
+        Optimobile.trackEvent(
+            eventType: OptimobileEvent.MESSAGE_OPENED,
+            properties: [
+                "type" : KS_MESSAGE_TYPE_PUSH,
+                "id": notificationId
+            ],
+            immediateFlush: true
+        )
```
cgwyllie: Rather than duplicating logic here, would it be preferable to call into the existing open tracking method: https://github.com/optimove-tech/Optimove-SDK-iOS/blob/259ed73cb8b1816dce85f4baee8b435418f258c1/OptimoveSDK/Sources/Classes/Optimobile/Optimobile%2BPush.swift#L212

I think if we change the contract to take in the `userInfo` aps payload & construct a model here, it's more flexible for the future and removes the need for the consumer to learn how to parse our ID out of the payload.

WDYT? :)

eligutovsky: Sure, `userInfo` is a typical data structure for notifications, and it might be useful here.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
```diff
@@ -209,27 +209,22 @@ extension Optimobile {
...
+            """)
             return
         }
-
         let params = ["type": KS_MESSAGE_TYPE_PUSH, "id": notification.id]
         Optimobile.trackEvent(eventType: OptimobileEvent.MESSAGE_OPENED, properties:params)
     }
 
-    @available(iOS 9.0, *)
-    internal func pushHandleOpen(withUserInfo: [AnyHashable: Any]?) {
```
cgwyllie, line 222: Is it safe to make these args non-optional?

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
```diff
@@ -209,27 +209,22 @@ extension Optimobile {
         Parameters:
             - notification: The notification which triggered the action
     */
-    static func pushTrackOpen(notification: PushNotification?) {
-        guard let notification = notification else {
+    static func pushTrackOpen(notification: PushNotification) {
+        if notification.id == 0 {
+            Logger.error("""
```
cgwyllie: Is this always an error?

For example, we may be passed a notification from another push provider, and the ID is (I think) expected to be 0. (unless the reason for optionals was if the initialiser of the PushNotification model is optional?)

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
cgwyllie: Also, presume this method has no consumers left as we don't support iOS 9 now? (so no swizzles call to this from app-delegate methods?)

eligutovsky: By default, this method is marked as internal and not exposed to the public. If someone tries to pass 'nil', the compiler will throw an error.

cgwyllie: Cool so long as everything changed is internal and it won't be breaking, seems fine.

Was just curious why we had them mostly as optional, if there's a path where we can call into it without a valid model or something.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
eligutovsky: Currently, the initialization of the PushNotification model returns a non-optional object with default values, where the `id` is set to 0. However, the check `notification.id == 0` is not accurate for checking Optimove payload. To improve this, we should apply the `Swift.Decodable` protocol or throw the corresponding error as a sign of invalid payload. 

For now, I can leave it as it is or apply `Swift.Decodable` to `PushNotification`. Either way, the error handling will remain the same with logging an error.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+Push.swift
eligutovsky: > Also, presume this method has no consumers left as we don't support iOS 9 now? (so no swizzles call to this from app-delegate methods?)

It is not in use in other parts of the code.

cgwyllie (approved):

PR #82, opened 2023-11-19
feat(API): delayed configuration

eligutovsky:
### Description of Changes

This pull request introduces the delayed initialization feature. 

#### Public changes

The update includes two new public APIs: 

1. `OptimoveConfigBuilder(region: Region, features: [Feature])` - this is new initializing the configuration builder. 
2. `Optimove.setCredentials(optimoveCredentials: String?, optimobileCredentials: String?)` - this is used for delayed setup credentials. 

Usage example of the delayed init feature:
```swift
// The new initializer
let config = OptimoveConfigBuilder(region: .US, features: [.optimove, .optimobile]).build()
Optimove.initialize(with: config)

// ... after some time ...

// The SDK will initialize only instances corresponding to passed credentials.
Optimove.setCredentials(optimoveCredentials: "OPTIMOVE_TOKEN", optimobileCredentials: "OPTIMOBILE_TOKEN")
```

It's important to note that the SDKs will not be initialized if there's no valid configuration. 

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [ ] `OptimoveCore.podspec`
- [ ] `OptimoveSDK.podspec`

- [ ] `README.md`
- [x] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [x] Init SDK with only T&T credentials
- [x] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


OptimobileShared/Credentials.swift
```diff
@@ -0,0 +1,8 @@
+//  Copyright © 2023 Optimove. All rights reserved.
+
+import Foundation
+
+struct Credentials: Codable {
```
eligutovsky: Rename to `OptimobileCredentials`

OptimobileShared/Network/NetworkClient.swift
```diff
@@ -73,10 +75,17 @@ class KSHttpClient {
...
-    @discardableResult func sendRequest(_ method: KSHttpMethod, toPath: String, data: Any?, onSuccess: @escaping KSHttpSuccessBlock, onFailure: @escaping KSHttpFailureBlock) -> URLSessionDataTask {
-        let request = newRequestToPath(toPath, method: method, body: data)
+    func sendRequest(_ method: KSHttpMethod, toPath: String, data: Any?, onSuccess: @escaping KSHttpSuccessBlock, onFailure: @escaping KSHttpFailureBlock) {
+        var request = newRequestToPath(toPath, method: method, body: data)
 
-        return sendRequest(request: request, onSuccess: onSuccess, onFailure: onFailure)
+        do {
+            try authorization.authorizeRequest(&request)
+        } catch {
+            onFailure(nil, error, nil)
```
eligutovsky: If the http_client encounters an error during authorization of the request, it passes that error to the `onFailure` block.

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -3,6 +3,18 @@
 import Foundation
 
 public struct OptimoveConfig {
+    /// Optimove initialization strategy.
+    struct InitializationStrategy: OptionSet {
```
vvoicehovics: strategy is different from the android one

1. in android the goal of it is to tell which credentials will be eventually given. If it's a T&T client we dont need to initialize optimobile ever -- it would just needlessly keep caching events and initing other irrelevant stuff. So, we given it optimove_only meaning that 'later we will give only optimove creds and now we dont want optimobile init'
2. in ios it's currently not used, .delayed is not telling which creds we are planning to receive, optimobile is always initialised

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -52,16 +52,24 @@ typealias Logger = OptimoveCore.Logger
...
-                Optimobile.initialize(config: optimobileConfig, initialVisitorId: visitorId, initialUserId: userId)
             }
         }
     }
+
+    /// Set the credentials for the Optimove server. Intent to use as a step for the delayed initialization.
+    public static func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?) {
+        let builder = OptimoveConfigBuilder(optimoveCredentials: optimoveCredentials, optimobileCredentials: optimobileCredentials)
+        let config = builder.build()
+        initialize(with: config)
```
vvoicehovics, line 76: is it true that  Optimobile.initialize is called twice (here and isOptimobileConfigured is always true)? This means

1. 2nd time wont succeed as there is a check if instance is already created
2. some things like sendDeviceInformation, maybeTrackPushDismissedEvents would be also called twice -- would get double event. Also some other helpers etc init twice, can it break smth?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -161,100 +189,155 @@ open class OptimoveConfigBuilder: NSObject {
...
 
-    private static func parseOptimoveCredentials(creds: String?) -> (tenantToken: String, configName: String)? {
-        guard let creds = creds,
-              let tuple = parseTuple(creds: creds),
-              let ver = tuple[0] as? Int
-        else {
-            return nil
+public extension OptimobileConfig {
+    enum Region: String, CaseIterable {
+        case EU1 = "eu-central-1"
```
vvoicehovics: i think dont need EU1 -- it's not used for optimove tenants and would needlessly confuse people

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -161,100 +189,155 @@ open class OptimoveConfigBuilder: NSObject {
         return self
     }
 
-    /**
-     Internal SDK embedding API, do not call or depend on this method in your app
-     */
-    @discardableResult public func setBaseUrlMapping(baseUrlMap: ServiceUrlMap) -> OptimoveConfigBuilder {
```
vvoicehovics, line 167: are we sure no hybrid sdk use this? @cgwyllie you probably remember what was this for?

cgwyllie: A good question, I'm not sure what it was used by, but certainly stuff we annotated as embedding API tends to be hybrid SDKs...

Looks like Unity needs it: https://github.com/search?q=org:optimove-tech++setBaseUrlMapping&type=code

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
cgwyllie: If we're opening these up as well, we probably should name them consistently with Optimove (e.g. 'EU', 'US', 'DEV'), and values can be the 'internal' value we need?

OptimobileShared/Network/NetworkClient.swift
```diff
@@ -53,14 +61,8 @@ class KSHttpClient {
...
-    func setBasicAuth(user: String, password: String) {
-        let creds = "\(user):\(password)"
-        let data = creds.data(using: .utf8)
-        let base64Creds = data?.base64EncodedString()
-
-        if let encoded = base64Creds {
-            authHeader = "Basic \(encoded)"
-        }
+    deinit {
+        invalidateSessionCancellingTasks(true)
```
vvoicehovics: AnalyticsHelper has `invalidateSessionCancellingTasks(false)`. 

now every HttpClient does this. Does it mean invalidateSessionCancellingTasks is going to be called 2 times with different values?

OptimobileShared/Network/NetworkClient.swift
vvoicehovics: can we go through what happens to events that failed to sync due to missing header. When will we attempt to sync them again?

it seems to work different VS android
- in android if it fails WorkManager will do a retry with default exponential strategy
- in ios sync happens when AnalyticsHelper is initialised and when app goes to bg (tracking immediate bg event). So, if we did 2nd init seems they would instantly sync --> seems ok

In general, for android and ios we should settle on whether we want flush when creds given or OK to rely on later retry

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
vvoicehovics: does it need it though? this seems to be just refs from included sdks

eligutovsky: You're right -- this method was in use by other SDKs. Moving it back

OptimobileShared/Network/NetworkClient.swift
eligutovsky: Good catch. I didn't notice it. Move this function with the correct argument to a holding object

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
eligutovsky: @vvoicehovics the ref in Unity SDK was only to iOS SDK xcframework. We can delete this API.

OptimoveSDK/Sources/Classes/Optimove.swift
eligutovsky: Change the initialization logic of `Optimobile`. Now it consumes the config with feature flags and can differentiate inits.

OptimobileShared/Network/NetworkClient.swift
eligutovsky: I added a new system notification named `optimobileInializationFinished`, where ddl and analytic services are subscribers. They will try to send a cache to server

OptimobileShared/MediaHelper.swift
```diff
@@ -2,17 +2,18 @@
 
 import Foundation
 
-public enum MediaHelper {
-    private static let mediaResizerBaseUrl: String = "https://i.app.delivery"
+enum MediaHelper {
+    /// Use ``Region.US`` as fallback region.
+    static let mediaResizerBaseUrl: String = "https://i-us-east-1.app.delivery"
```
vvoicehovics, line 7: huh, why, is it not better to crash? (why falling back to us)

eligutovsky: It is crucial that we avoid any crashes at all costs! 😅

This is a fallback URL that will only be used if there is no region value stored. However, in most cases, the region value should be stored as it is required for the SDK initialization process.

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
```diff
@@ -208,40 +207,23 @@ public class OptimoveNotificationService {
...
-    fileprivate class func initializeAnalyticsHelper() {
-        let apiKey = KeyValPersistenceHelper.object(forKey: OptimobileUserDefaultsKey.API_KEY.rawValue)
-        let secretKey = KeyValPersistenceHelper.object(forKey: OptimobileUserDefaultsKey.SECRET_KEY.rawValue)
-
-        guard let apiKey = apiKey as? String, let secretKey = secretKey as? String else {
-            print("Extension: authorization credentials not present")
-            return
+        do {
+            let pendingAnalytics = PendingAnalyticsImpl(storage: KeyValPersistenceHelper.self)
+            try pendingAnalytics.add(
```
vvoicehovics: when is this flushed? is it slightly changing behaviour, e.g. can it happen that message is delivered and then user never opens app , and we never receive the event? 


still have 50 files to read, but is it flushed only when initialisation finishes? what if it's standard init?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -35,12 +37,49 @@ class AnalyticsHelper {
...
+            let metrics = try pendingAnalytics.readAll()
+            Logger.debug("Found \(metrics.count) pending metrics")
+            metrics.forEach { metric in
+                self.trackEvent(
+                    eventType: metric.eventType.rawValue,
+                    atTime: Date(timeIntervalSince1970: Double(metric.timestamp)),
+                    properties: metric.properties,
+                    immediateFlush: true,
+                    onSyncComplete: { _ in
+                        removeAllPendingMetrics()
```
vvoicehovics: we track event for each metric, but removeAll after each one completes. it's probably minor and not important enough but theoretically could miss events if
- 1st succeeds, rest fail
- receive new event while syncing these

OptimoveSDK/Sources/Classes/Optimobile/Network/UrlBuilder.swift
```diff
@@ -0,0 +1,63 @@
...
+            throw Error.regionNotSet
+        }
+    }
+
+    required init(storage: KeyValPersistenceHelper.Type) {
+        self.storage = storage
+    }
+
+    func urlForService(_ service: Service) throws -> URL {
+        let baseUrlMap = try UrlBuilder.defaultMapping(for: region)
```
vvoicehovics: always uses default mapping? e.g. what if setBaseUrlMapping is used

OptimoveSDK/Sources/Classes/Optimove.swift
vvoicehovics: it uses new builder here. Does it mean all initial configuration options (like in-app strategy) are lost when we give credentials?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
vvoicehovics: turns used by in-app people, shall put back

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -161,100 +217,150 @@ open class OptimoveConfigBuilder: NSObject {
...
-                isRelease: _isRelease
-            )
-        }
+        let tenantInfo: OptimoveTenantInfo? = {
+            if let _tenantToken = _tenantToken,
+               let _configName = _configName
+            {
+                return OptimoveTenantInfo(tenantToken: _tenantToken, configName: _configName)
+            }
+            Logger.info("\(OptimoveTenantInfo.self) building failed.")
```
vvoicehovics: mb it's more like 'skipped' Vs 'failed'. optimove creds can be null . same for optimobile below

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
eligutovsky: You're right. Fixed

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
```diff
@@ -89,12 +77,27 @@ class Optimobile {
...
-        if instance !== nil {
-            assertionFailure("The OptimobileSDK has already been initialized")
+    static func initialize(config optimoveConfig: OptimoveConfig, initialVisitorId: String, initialUserId: String?) {
+        if instance !== nil, optimoveConfig.features.contains(.delayedConfiguration) {
+            completeDelayedConfiguration(config: optimoveConfig.optimobileConfig!)
+            return
+        }
+
+        guard instance == nil else {
+            Logger.error("The OptimobileSDK has already been initialized")
```
vvoicehovics: throw here? previous intention was such

OptimoveCore/Sources/Classes/Storage/FileStorage.swift
```diff
@@ -7,56 +7,105 @@ public protocol FileStorage {
...
     /// - Parameters:
     ///   - fileName: The file name on disk space.
-    func delete(fileName: String) throws
+    ///   - isTemporary: set `true` if the requested file should be saved in a temporary directory.
+    ///   The file will be removed after app will be terminated. Default value is `false`.
+    func delete(fileName: String, isTemporary: Bool) throws
+}
+
+public extension FileStorage {
+    func isExist(fileName: String) -> Bool {
```
eligutovsky, line 64: `isTemporary: Bool = false`

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
```diff
@@ -116,6 +119,28 @@ class Optimobile {
         maybeAlignUserAssociation(initialUserId: initialUserId)
     }
 
+    static func completeDelayedConfiguration(config: OptimobileConfig) {
+        guard let credentials = config.credentials else {
+            Logger.warn("No credentials provided for delayed configuration completion.")
```
vvoicehovics: also here and everywhere when smth is unexpected in android we throw

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
```diff
@@ -116,6 +119,28 @@ class Optimobile {
...
+        updateStorageValues(config)
+        setCredentials(credentials)
+        NotificationCenter.default.post(name: .optimobileInializationFinished, object: nil)
+    }
+
+    static func setCredentials(_ credentials: Credentials) {
+        Optimobile.instance?.credentials = credentials
+    }
+
+    static func updateStorageValues(_ config: OptimobileConfig) {
```
vvoicehovics, line 156: do we need to update region? in android we dont because region shouldnt change / we already set it in the new builder init

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
```diff
@@ -116,6 +119,28 @@ class Optimobile {
...
 
+    static func completeDelayedConfiguration(config: OptimobileConfig) {
+        guard let credentials = config.credentials else {
+            Logger.warn("No credentials provided for delayed configuration completion.")
+            return
+        }
+        Logger.info("Completing delayed configuration with credentials: \(credentials)")
+        updateStorageValues(config)
+        setCredentials(credentials)
+        NotificationCenter.default.post(name: .optimobileInializationFinished, object: nil)
```
vvoicehovics, line 149: another thing android does is syncing in-apps

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: The region from init and credentials potentially can be different. I think we need to use the latest one to prevent unauthorized errors.

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -46,20 +46,40 @@ typealias Logger = OptimoveCore.Logger
...
+    public static func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?) {
+        let builder = OptimoveConfigBuilder(optimoveCredentials: optimoveCredentials, optimobileCredentials: optimobileCredentials)
+        builder.setFeatures([builder.features, .delayedConfiguration])
+        let config = builder.build()
+        initialize(with: config)
+    }
 
-                Optimobile.initialize(config: optimobileConfig, initialVisitorId: visitorId, initialUserId: userId)
-            }
+    public static func isFeatureRunning(_ feature: Feature) -> Bool {
```
vvoicehovics, line 79: public api isFeatureRunning? dont see it used anywhere

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: I can throw an error here and then catch it in a caller, but there I will use the same logger. We should catch our errors and do not crash the host app.

Previous `assertionFailure` stops in debug only. Maybe we should do it additionally to logger.
https://developer.apple.com/documentation/swift/assertionfailure(_:file:line:)#discussion

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -46,20 +46,40 @@ typealias Logger = OptimoveCore.Logger
...
                 }
+            }
+        }
+    }
 
-                let userId = try? serviceLocator.storage().getCustomerID()
+    /// Set the credentials for the Optimove server. Intent to use as a step for the delayed initialization.
+    public static func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?) {
+        let builder = OptimoveConfigBuilder(optimoveCredentials: optimoveCredentials, optimobileCredentials: optimobileCredentials)
+        builder.setFeatures([builder.features, .delayedConfiguration])
```
vvoicehovics: another difference here is that in android we dont update features after initial initialisation. If you initially say you want optimobile only and then give optimove credentials/dont give optimobile -- we throw. That's 1 of goals of FeatureSet -- telling early what do they want so that later can make sure they gave credentials for that

vvoicehovics: and so, perhaps fix for everything is to set credentials onto config, not builder . (at least that's what's in android)

OptimobileShared/MediaHelper.swift
vvoicehovics: it will crash anyway (hopefully) if you try to request wrong region , just it will be more obscure. If it doesnt crash we will just be silently not sending any events

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
vvoicehovics: k, in android it also doesnt crash here, so, let's not. But i think crashing host app is sometimes fine/better than silently not doing what's expected

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
vvoicehovics: my main assumption (reasonable?) was that region doesnt change, and if it does -- it's an error. 

Otherwise why do we even provide region to the builder? If region can change based on brand -- this means we wouldnt know it at start

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: let's keep `assertionFailure` here -- it will crash in the debug mode only. But keep a production version crash-free.

OptimobileShared/MediaHelper.swift
eligutovsky: It is a bad way to produce a crash in notification_service_extension. In case of a crash or a timeout limit -- a notification will be shown without an image by the OS.
So, in the worst case, there will be no image attached to the notification.

Ideally, the URL should be placed in the payload.

OptimoveSDK/Sources/Classes/Optimove.swift
eligutovsky: It used in the QA app

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
eligutovsky: These metrics are not publicly available. 
We will not get these metrics if a user never opens the app.

eligutovsky: Add notification for standard init

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: Good question why do we need to pass the region on init. As I remember it needs only for notification images.

eligutovsky: What if regions in a set of tenants' credentials differ? It seems like an artificial obstacle, follow the origin on init.

OptimobileShared/MediaHelper.swift
vvoicehovics: mmm, ok, mb crashing is bad. But i think silently displaying notification without picture (i.e.not how they wanted it) is also quite bad. It's questionable what's better -- display without pic or not display at all.

But at least could add some logging? + still default US seems weird. Is it to make it work sometimes? we could throw and catch

vvoicehovics: ok, to speed up the very-wanted release, let's fallback to US, not major enough

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
vvoicehovics: could there be thread safety issues?  when we remove 1 we read all and set all. 

however, since delivered events are tech debt, let's deal later

eligutovsky: it iterates a copy of all metrics. i do not see a race condition.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
vvoicehovics: yes, it's only for picture. You are right that for picture giving region in notification payload could be an option (especially if we considered it at start and havent rolled android already). 

still
1. region would have to be part of credentials anyway. Loading notification picture is a single use case covered by region in payload. Hypothetically there could be another use case where we need region before finishing init
2. so far all tenants were in the same region 

then it would depend which one of 1. or 2. is more likely / sooner to change.

anyway, fine to keep here to speed up

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
vvoicehovics: remove delivered events for now together with PendingAnalytics

OptimoveSDK/Sources/Classes/Optimove.swift
eligutovsky: This part was changed: added a new OptimoveConfigBuilder init with for the current config -- it saves all existed settings.

OptimoveSDK/Sources/Classes/Optimove.swift
vvoicehovics: there is some work to not copy each key individually back into builder from config. They are very easy to forget adding and very hard to spot. And if miss -- config after finishing init wont be the same as. created by users

tech debt

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -278,9 +278,11 @@ class InAppPresenter: NSObject, WKScriptMessageHandler, WKNavigationDelegate {
         #else
             let cachePolicy = URLRequest.CachePolicy.reloadIgnoringCacheData
         #endif
-        let url = Optimobile.getInstance().urlBuilder.urlForService(.iar)
-        let request = URLRequest(url: URL(string: url)!, cachePolicy: cachePolicy, timeoutInterval: 8)
-        webView.load(request)
+        if let urlString = KeyValPersistenceHelper.object(forKey: OptimobileUserDefaultsKey.IAR_BASE_URL.rawValue) as? String,
```
cgwyllie: Is there a reason we persist IAR base URL? It is constant & non-regionalised for production.

(and sorry, appreciate I'm late to the party, there will be context I'm missing, just curious--happy to discuss next week!)

eligutovsky: Thank you for your comment. 
Indeed, the IAR_BASE_URL is a constant and we can skip passing it through the storage layer.
Removed from storage keys and used as a constant in in-app presenter.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: Added `sync` invocation in the in-app for `.optimobileInializationFinished` notification.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: It was converted to an error throw. Check `Optimobile.Error`

Makefile
```diff
@@ -0,0 +1,31 @@
+XCODE ?= 15.0.1
+
+export TEST_DESTINATION ?= platform=iOS Simulator,OS=17.0,name=iPhone 15
+export DEVELOPER_DIR = $(shell bash ./scripts/get_xcode_path.sh ${XCODE} $(XCODE_PATH))
+export CURRENT_PROJECT_VERSION = 5.6.0
```
cgwyllie: Is this SDK version?

Should we update the checklist in the PR template to mention bumping this?

OptimobileShared/Credentials.swift
cgwyllie: Did we choose not to rename this?

OptimobileShared/MediaHelper.swift
```diff
@@ -2,17 +2,18 @@
...
 
     static func getCompletePictureUrl(pictureUrl: String, width: UInt) -> URL? {
-        if (((pictureUrl as NSString).substring(with: NSRange(location: 0, length: 8))) == "https://") || (((pictureUrl as NSString).substring(with: NSRange(location: 0, length: 7))) == "http://") {
+        if pictureUrl.hasPrefix("https://") || pictureUrl.hasPrefix("http://") {
```
cgwyllie, line 10: Nice :)

OptimobileShared/MediaHelper.swift
cgwyllie: For this, agree in the sense having a default that _may_ work sometimes is not perhaps the best.

It could obscure setup problems for US tenants which may affect other things if not correctly configured.

Personally think it is better if the region is missing to show the content without the picture & log a warning about bad init state.

In line with Vlad tho, we may have to ship like this and create a tech debt to discuss.

OptimoveSDK/Sources/Classes/Legacy/Logger/OptiLoggerOutputStream.swift
```diff
@@ -3,7 +3,7 @@
 import Foundation
 
 /// Receiver of logs of OptimoveSDK.
-@objc public protocol OptiLoggerOutputStream {
+public protocol OptiLoggerOutputStream {
```
cgwyllie, line 6: Any. reason we're removing objc annotations here? (presume no direct consumers so it's not necessary)

Package.swift
```diff
@@ -23,6 +23,9 @@ let package = Package(
             targets: ["OptimoveNotificationServiceExtension"]
         ),
     ],
+    dependencies: [
+        .package(url: "https://github.com/WeTransfer/Mocker", from: "3.0.1"),
```
cgwyllie, line 27: Is this dep. only for tests / should it be declared at the 'top' level?

(I'm not so familiar w/ SPM so this may not be an accurate understanding)

OptimoveSDK/Sources/Classes/Legacy/Logger/OptiLoggerOutputStream.swift
eligutovsky: I am trying to remember the exact reason for removing it, but cannot. In any case, it is used for QA app access for internal logs -- it does not affect SDK usage for tenants.

Shared/Sources/Fixture/GlobalConfigFixture.swift
```diff
@@ -3,35 +3,35 @@
 import Foundation
 @testable import OptimoveCore
 
-final class GlobalConfigFixture: FileAccessible {
-    let fileName: String = "core_events.json"
+public final class GlobalConfigFixture: FileAccessible {
```
cgwyllie, line 6: What's the reason for lifting these classes/methods out of Tests, do we now use them in the SDK?

If not, do we 'pollute' the public classes in our SDK modules with test helpers?

(just want to understand, ofc may have the wrong understanding!)

project.yml
```diff
@@ -1,147 +1,147 @@
-name: Optimove
-options:
-  deploymentTarget:
-    iOS: "12"
+# name: Optimove
```
cgwyllie: Is this no longer used / should we remove?

Package.swift
eligutovsky: The SPM package file has a very strict structure -- moving dep into another position will show an error.

Shared/Sources/Fixture/GlobalConfigFixture.swift
eligutovsky: Checked it, and it is a part of `OptimoveTest` library that is used only for testing purposes.

project.yml
eligutovsky: Due to a shortage of time to prepare the file for building the correct test schema, I ended up creating a xcodeproj for tests. Although I might revert to this format in the future, this approach eliminates the need for an xcodeproject, which typically has a large number of lines in a PR when the project structure is changed.

Removed file.

Makefile
eligutovsky: Removed this line from here as not relevant.

OptimobileShared/MediaHelper.swift
eligutovsky: Added a tech debt task

OptimobileShared/Credentials.swift
eligutovsky: Renamed to `OptimobileCredentials`

OptimoveCore/Sources/Classes/Storage/StorageFacade.swift
```diff
@@ -99,6 +101,29 @@ extension UserDefaults: KeyValueStorage {
     }
 }
 
+public final class InMemoryStorage: KeyValueStorage {
+    private var storage = [StorageKey: Any]()
```
cgwyllie, line 105: Are there any thread-safety concerns w/ this in-memory store?

(not familiar with how it's being used, or even if stale reads could matter)

OptimoveCore/Tests/Sources/Storage/OptimoveStorageFacadeTests.swift
```diff
@@ -37,4 +38,25 @@ class OptimoveStorageFacadeTests: XCTestCase {
         let value: String? = storage[key]
         XCTAssert(value == stub_string)
     }
+
+    func test_try_get_subscript() throws {
```
cgwyllie, line 42: Are these tests to cover that we use the alternate storage for certain keys?

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
```diff
@@ -208,42 +206,6 @@ public class OptimoveNotificationService {
         KeyValPersistenceHelper.set(newBadge, forKey: OptimobileUserDefaultsKey.BADGE_COUNT.rawValue)
     }
 
-    fileprivate class func trackDeliveredEvent(dispatchGroup: DispatchGroup, userInfo: [AnyHashable: Any], notificationId: Int) {
```
cgwyllie, line 211: Where did this logic move to, did we lose it?

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
```diff
@@ -29,11 +30,21 @@ class InAppManager {
...
-    init(_ config: OptimobileConfig) {
+    init(_ config: OptimobileConfig, httpClient: KSHttpClient) {
+        self.httpClient = httpClient
         presenter = InAppPresenter(displayMode: config.inAppDefaultDisplayMode)
         syncQueue = DispatchQueue(label: "kumulos.in-app.sync")
+
+        finishedInitializationToken = NotificationCenter.default
+            .addObserver(forName: .optimobileInializationFinished, object: nil, queue: nil) { [weak self] notification in
+                guard let self = self else { return }
+                self.sync()
```
cgwyllie: Is it correct to call `sync()` here and not `syncDebounced()`?

If we directly call `sync()` we will poll the server on every initialisation which is not the previous behaviour and may result in far more requests for the sync endpoint.

LMK if you want to discuss :)

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
```diff
@@ -14,7 +14,11 @@ enum InAppAction: String {
     case REQUEST_RATING = "requestAppStoreRating"
 }
 
-class InAppPresenter: NSObject, WKScriptMessageHandler, WKNavigationDelegate {
+final class InAppPresenter: NSObject, WKScriptMessageHandler, WKNavigationDelegate {
+    enum Constants {
+        static let baseUrl = "https://iar.app.delivery"
```
cgwyllie: I know we discussed this being constant in the sense it doesn't need written to storage, but should it not still come from `urlBuilder.urlForService(.iar)` or equivalent?

I believe the in-app team are currently working on an E2E local dev environment that enables baseUrl for this to be pointed at a local renderer instance from `setUrlMapping(...)`.

WDYT?

OptimoveSDK/Sources/Classes/Optimobile/Network/AuthorizationMediator.swift
```diff
@@ -0,0 +1,67 @@
...
+    func authorizeRequest(_: inout URLRequest) throws
+}
+
+typealias CredentialsProvider = () -> OptimobileCredentials?
+
+final class AuthorizationMediator {
+    var basicAuthorization: String?
+    let provider: CredentialsProvider
+
+    init(basicAuthorization: String? = nil, provider: @escaping CredentialsProvider) {
```
cgwyllie: Do we need to take auth in at init, should we prefer calling setBasicAuth at consumers who pass during init instead?

(this would centralise logic of encoding the values in one place)

I haven't seen all usage sites yet so maybe not appropriate!

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
```diff
@@ -22,11 +22,23 @@ public enum InAppDisplayMode: String {
...
+        case alreadyInitialized
+        case configurationIsMissing
+        case noCredentialsProvidedForDelayedConfigurationCompletion
+
+        var errorDescription: String? {
+            switch self {
+            case .alreadyInitialized:
+                return "The OptimobileSDK has already been initialized."
+            case .configurationIsMissing:
+                return "OptimoveConfig is missing."
```
cgwyllie: Optimove or Optimobile?

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+DeepLinking.swift
```diff
@@ -37,26 +37,42 @@ public enum DeepLinkResolution {
...
-    init(_ config: OptimobileConfig, urlBuilder: UrlBuilder) {
+    init(_ config: OptimobileConfig, httpClient: KSHttpClient) {
         self.config = config
-        httpClient = KSHttpClient(
-            baseUrl: URL(string: urlBuilder.urlForService(.ddl))!,
-            requestFormat: .rawData,
-            responseFormat: .rawData,
-            additionalHeaders: [
-                "Content-Type": "application/json",
-                "Accept": "appliction/json",
```
cgwyllie, line 55: Do we lose these headers?

I think certainly the Accept will be required to receive correct error responses from the server in some cases.

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -357,13 +381,13 @@ public extension Optimove {
     @objc func trackIBeaconProximity(beacon: CLBeacon) {
         Optimobile.trackIBeaconProximity(beacon: beacon)
     }
-    
+
     /// Records a notification open event
     /// - Parameter userInfo - The userInfo dictionary you received in the push notification payload
-    @objc public func trackOpenMetric(userInfo: [AnyHashable: Any]) {
```
cgwyllie, line 363: Does this inherit the class `public` access level by default?

eligutovsky: The default access level by default is internal. We have to explicitly define it as public here.

OptimoveSDK/Sources/Classes/Optimobile/Optimobile+DeepLinking.swift
eligutovsky: This part moved to https://github.com/optimove-tech/Optimove-SDK-iOS/blob/a715d8d210e98e92bf6802f7fb211bb7bc23a22b/OptimoveSDK/Sources/Classes/Optimobile/Network/NetworkClient.swift#L54

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: You're right -- it is typo.

OptimoveCore/Tests/Sources/Storage/OptimoveStorageFacadeTests.swift
eligutovsky: No, this test checking a new swift subscript method which can throws an error. It was not possible a few years ago with previous versions of swift lang

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
eligutovsky: We decided to remove this collecting `delivered` metric. These metrics are not public.

OptimoveCore/Sources/Classes/Storage/StorageFacade.swift
eligutovsky: Adding an access queue for being thread-safety.

It's worth refactoring the storage object to work with `async/await` and use `Actor` type for thread-safety.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
eligutovsky: Not sure about using `syncDebounced()` here. The idea is to keep the same behaviour in delayed init as it was for regular init. 
Replaced `sync()` with `handleEnrollmentAndSyncSetup()`.

OptimoveSDK/Sources/Classes/Optimobile/Network/AuthorizationMediator.swift
eligutovsky: The `basicAuthorization` was not in use here. Removed.

Added the `AuthorizationStrategy` as an argument for `HttpAuthorizationProtocol`.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift
eligutovsky: Reverting the usage of `UrlBuilder` here

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
vvoicehovics: why with `handleEnrollmentAndSyncSetup`? -- it does stuff like adding observer and must have been already called during initial partial init . Is it because ordinary sync wouldnt present anything?

syncing here is kinda for in-apps with tickles:
1. kill app
2. receive push
3. open push, app launched ... loading credentials
4. setCredentials ---> now we want to display in-apps that would show if init wasnt delayed

sync debounced with 1h between syncs wouldnt ensure it either

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
vvoicehovics: *which was another temp thing to move this forward, we do report this metric on android (yes, not used currently)

OptimoveCore/Sources/Classes/Storage/StorageFacade.swift
vvoicehovics: this in memory storage was for delivered events, right, currently not used?

eligutovsky: > this in memory storage was for delivered events, right, currently not used?

No, we stopped collecting delivery metrics.
 
The in-memory storage used by the optimove part, keeps `[.tenantToken, .version]` in memory, where `version` is a config name from the optimove config.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
eligutovsky: The `optimobileInializationFinished` notification triggers the previous regular init flow also for the delayed initialization, simplifying these two cases into one init flow.

eligutovsky: > Is it because ordinary sync wouldnt present anything?

It will do the intended work.

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
cgwyllie: OK have we recorded the appropriate tech debts here / raised with all relevant stakeholders? (whichever direction if we plan to remove for all or fix iOS)

project.yml
cgwyllie: Cool.

Whilst xcodeproject can result in large numbers of changed lines on a diff, it's treated like a generated file, so not sure it's a strong reason not to have one (we also generally don't broadly change structure too often (I would hope ;)).

Shared/Sources/Fixture/GlobalConfigFixture.swift
cgwyllie: Cool so the classes aren't accessible in the SDK imports?

OptimoveSDK/Sources/Classes/Optimove.swift
cgwyllie: Hm yes, so have we accidentally made this method internal with this change?

eligutovsky: Oh, this line is located in `public extension Optimove` -- it is `public` by the extension definition.

OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
eligutovsky: > OK have we recorded the appropriate tech debts here / raised with all relevant stakeholders? (whichever direction if we plan to remove for all or fix iOS)

Created a tech debt and a request to know if this metric is still relevant.

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
eligutovsky: > why with `handleEnrollmentAndSyncSetup`? -- it does stuff like adding observer and must have been already called during initial partial init . Is it because ordinary sync wouldnt present anything?
> 
> syncing here is kinda for in-apps with tickles:
> 
> 1. kill app
> 2. receive push
> 3. open push, app launched ... loading credentials
> 4. setCredentials ---> now we want to display in-apps that would show if init wasnt delayed
> 
> sync debounced with 1h between syncs wouldnt ensure it either

Based on my comprehension of the current code, it seems adequate to initialize only the database during the initiation process. Any actions related to adding an observer on `appBecameActive` can be deferred to the final init stage since it involves synchronization with the server. Additionally, there is no need to perform synchronization with the server in the absence of credentials.

Shared/Sources/Fixture/GlobalConfigFixture.swift
eligutovsky: That is correct.
The `OptimoveTest` is available only for test targets `OptimoveSDKTests` and `OptimoveCoreTests`, according `Package.swift`

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
vvoicehovics: we keep introducing subtle differences with previous and with android:

1. now every time non-delayed init finishes we fire the event, so, we trigger all observer actions: flush events, check cached links etc. In android non-delayed init proceeds exactly as before. These extra actions are essentially only needed for delayed init
2. i) Can calling `updateUserConsent` when init finishes affect inbox which should work without credentials? Because user consent is not set until init is finished. Practically now no. ii) IN_APP_CONSENT_CHANGED event timestamp now becomes time when we got credentials and finished init. But essentially updating consent doesnt depend on credentials, we do it when we know it should be updated

This new behaviour might be still ok, but what i dont like most is that it starts diverging from android here and there. Biggest Q here is flushing events every time normal init finishes

OptimoveSDK/Sources/Classes/Optimove.swift
cgwyllie: Ah yes, that's what I was asking! cool 😎

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -13,34 +13,41 @@ class KSEventModel: NSManagedObject {
...
 
         initContext()
 
-        DispatchQueue.global().async {
-            if self.migrationAnalyticsContext != nil {
-                self.syncEvents(context: self.migrationAnalyticsContext)
+        finishedInitializationToken = NotificationCenter.default
+            .addObserver(forName: .optimobileInializationFinished, object: nil, queue: nil) { [weak self] notification in
+                guard let self = self else { return }
+                self.flushEvents()
```
cgwyllie: Does it matter that we're changing thread semantics here?

OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppManager.swift
cgwyllie: Thanks Eli, I think `handleEnrollmentAndSyncSetup()` is indeed equivalent w.r.t. what the SDK previously did during init, so seems the correct choice here.

@vvoicehovics interesting question regarding inbox, at first I'm not sure it would matter, but perhaps we should discuss. Practically you can't see inbox items until they're synced anyway, and consent is cached, so this only affects first time init as best I can tell.

Regarding flushing of events, I don't believe this is different (at least on iOS) to how it was before, we always attempted a flush on SDK init here, so I'm less concerned about that.

Let's schedule some time to take a discussion around these topics and look to move this over the line :)

cgwyllie: (if we find inconsistencies as we go, we should capture under a tech feature perhaps to review)

vvoicehovics: oh, okay, i thought they flushed only when goes bg

you could have inbox items already in db, but consent lost in some way (in any case can clear storage not clearing db?). But y, consent is not used by inbox / unlikely

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
eligutovsky: It does not matter because the actual work will be done on CoreData's context thread.

```swift 
    private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
        context?.performAndWait {
            let results = fetchEventsBatch(context)
```

https://github.com/optimove-tech/Optimove-SDK-iOS/blob/8d6bccac9cbbf6ce76da4a857d0f84554dcadf1d/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift#L163C18-L163C18

OptimoveSDK/Sources/Classes/Optimobile/Optimobile.swift
eligutovsky: add `assertion`

vvoicehovics (approved):

cgwyllie (approved):

PR #97, opened 2023-12-14
refactor: json [7]

eligutovsky:
### Description of Changes

Use `GenericJSON` that conforms to `Codable`

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [ ] `OptimoveCore.podspec`
- [ ] `OptimoveSDK.podspec`

- [ ] `README.md`
- [ ] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


Package.swift
```diff
@@ -24,12 +24,14 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/WeTransfer/Mocker", from: "3.0.1"),
+        .package(url: "https://github.com/iwill/generic-json-swift", from: "2.0.2"),
```
cgwyllie, line 27: Hey @eligutovsky , have we had any discussion about adding new dependencies to the SDK?

Some clients are a little sensitive to the size/dependency footprint, and dependencies can make integration more tricky if there are conflicting versions.

Just a thought, we should consider it carefully.

OptimoveSDK/Sources/Classes/Optimobile/OptimoveInApp.swift
```diff
@@ -4,49 +4,53 @@ import CoreData
...
+final class InAppInboxItem {
     private(set) var id: Int64
     private(set) var title: String
     private(set) var subtitle: String
     private(set) var availableFrom: Date?
     private(set) var availableTo: Date?
     private(set) var dismissedAt: Date?
     private(set) var sentAt: Date
-    private(set) var data: NSDictionary?
+    private(set) var data: ObjcJSON?
```
cgwyllie, line 15: Are these 'breaking' API changes / do we expose/leak the type from a library to the consumer app?

Package.swift
eligutovsky: This dependency was copy-pasted to the source code previously. I will remove this dependency and will keep it copy-pasted.

cgwyllie: Sorry not sure I understand, you mean we already depend on it? (if so, fair enough, might propose the inverse of dependency simplification in the future, but that's another topic 😅)

eligutovsky: Yes, we're already depend on it. See `OptimoveSDK/Sources/Classes/JSON/JSON.swift` -- in removed files of this PR.

OptimoveSDK/Sources/Classes/Optimobile/OptimoveInApp.swift
eligutovsky: The `CoreData` entity model is only meant for internal use and should not be accessed by consumer apps.
This change has been made to fix the issue of `NSDictionary` not conforming to the `Codable` protocol. It is important to ensure that values stored in the `CoreData` are `Codable` before they are stored. This will help ensure that the data is ready to be send through network.

cgwyllie: Great, all for these kind of things in general, Codable will be simpler, but I think this model `InAppInboxItem` is not the Core Data entity, it is the model we expose to consumers, so changing this field could be breaking/leaky.

Package.swift
cgwyllie: Ah, clear!

Next question: do you know why we 'vendored' this library code into the SDK in the past, does it matter?

eligutovsky: It was added previously because the SDK allows to get `[String: Any]` as a parameter of an event. 

See https://github.com/optimove-tech/Optimove-SDK-iOS/pull/97#discussion_r1426850648

It might worth to replace `JSON` with `Data` by using `JSONSerialization`.

OptimoveSDK/Sources/Classes/Optimobile/OptimoveInApp.swift
eligutovsky: Hmm, this class cannot be exposed to a consumer -- it is not public

Package.swift
eligutovsky: Ideally, we need to decode any internal json to a Codable model.

PR #111, opened 2024-08-07
BA#218543 Add preference center functionality

daniela-opti:
### Description of Changes

Expose methods for getting and setting preferences 

Here are all the tests that I ran in the QA app to verify that preference center works  -> [Notion link](https://www.notion.so/All-of-my-test-cases-642348a5ebe9413fb336d29958bdf903)
### Breaking Changes

-   None

### Prepare:

- [x] Check `pod lib lint` passes
- [x] Update any relevant sections of the repository wiki pages on a branch

### Bump versions in:

- [x] `OptimoveCore.podspec`
- [x] `OptimoveSDK.podspec`

- [x] `README.md`
- [x] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenter.swift
```diff
@@ -0,0 +1,108 @@
...
+            print("Failed to create network request: \(error)")
+        }
+
+        throw NetworkError.requestFailed
+    }
+
+
+    func setPreferences(for customerId: String, brandGroupId: String, updates preferenceUpdates : [PreferenceUpdate]) async throws -> Preferences {
+        do {
+            // TODO: resolve this
```
kris-opti: do we want to keep this comment in the code?

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenterModels.swift
```diff
@@ -0,0 +1,91 @@
+import Foundation
```
kris-opti: Is this consistent where we have several classes / enums / structs all in one file?

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenter.swift
```diff
@@ -0,0 +1,106 @@
...
+            throw StorageError.noValue(.customerID)
+        }
+
+        return String(customerId)
+    }
+
+    func getPreferences(brandGroupId: String) async throws -> Preferences {
+        do {
+            // TODO: resolve region business
+            let region = "dev-pb"
```
scotthowie-optimove: this will just be "dev" with the updates to preference center endpoints

OptimoveCore/Sources/Classes/NetworkClient/NetworkRequest.swift
```diff
@@ -87,17 +87,32 @@ public extension HTTPHeader {
...
+        case textplain
+
+        var value: String {
+            switch self {
+            case .json:
+                "application/json"
+            case .tenantId(let id):
+                id
+            case .textplain:
+                "text/plain"
```
eligutovsky: We need to check what the old version of Swift Language does not support avoiding ‘return’ keyword. It can cause a compilation error.

OptimoveCore/Sources/Classes/NetworkClient/NetworkRequest.swift
```diff
@@ -87,17 +87,32 @@ public extension HTTPHeader {
...
     }
 
-    enum Values: String {
-        case json = "application/json"
+    enum Values {
+        case json
+        case tenantId(id: String)
+        case textplain
+
+        var value: String {
```
eligutovsky: Take a look at [CustomStringConvertable protocol](https://developer.apple.com/documentation/swift/customstringconvertible)

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -130,6 +130,66 @@ public extension Optimove {
...
+                        continuation.resume(throwing: error)
+                    }
+                }
+            }
+        }
+    }
+
+
+    @available(iOS 13.0, *)
+    @objc static func getPreferences(brandGroupId: String) async throws -> Preferences {
```
eligutovsky: I think objc does not support modern async. Need to be checked. 
This way, we might avoid using objc for async funcs

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -130,6 +130,66 @@ public extension Optimove {
...
+        do {
+            return try await shared.getPreferences(brandGroupId: brandGroupId)
+        } catch {
+            print(error)
```
eligutovsky: Forgotten print

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenter.swift
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    private func getCustomerId() throws -> String {
+        guard let customerId: String = storage[.customerID] else {
+            throw StorageError.noValue(.customerID)
+        }
+
+        return String(customerId)
+    }
```
eligutovsky: Let's try to move it to storage accessors.

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenter.swift
```diff
@@ -0,0 +1,106 @@
...
+    func getPreferences(brandGroupId: String) async throws -> Preferences {
+        do {
+            // TODO: resolve region business
+            let region = "dev-pb"
+            let tenantId = try getTenantId()
+            let customerId = try getCustomerId()
+
+            let request = NetworkRequest(
+                method: .get,
+                baseURL: URL(string: "https://preference-center-\(region).optimove.net")!,
```
eligutovsky: Potential runtime exception with using exclamation mark. Use try cast instead

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenter.swift
```diff
@@ -0,0 +1,106 @@
...
+            )
+
+
+            let data: Data = try cast(await networkClient.performAsync(request))
+            let preferences = try JSONDecoder().decode(Preferences.self, from: data)
+
+            return preferences
+        }
+        catch {
+            print("Failed to create network request: \(error)")
```
eligutovsky: Forgotten print

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenterModels.swift
```diff
@@ -0,0 +1,91 @@
...
+
+struct PreferenceCenterConfig: Codable {
+    var region: Region
+    var tenantId: Int
+    var brandGroupId: String
```
eligutovsky: Maybe Iam wrong, but I would consider to use let instead var in a struct.

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenterModels.swift
```diff
@@ -0,0 +1,91 @@
...
+}
+
+public class Preferences: NSObject, Codable {
```
eligutovsky: Consider using struct first. 
In this case, with exposing struct to objc, we need to wrap up the Preference struct with PrefernceObjc class. So, we will have doubled API. It will help us to use only modern lang features internally in sdk. 

The sdk should have some examples of it. I will find it

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenterModels.swift
```diff
@@ -0,0 +1,91 @@
...
+
+public class Topic: NSObject, Codable {
+    public var topicId: String
+    var topicName: String
+    var topicDescription: String
+    var channelSubscription: [Channel]
+
+    public func getId() -> String {
+        return topicId
+    }
```
eligutovsky: Instead of writing getters, Swift Lang provides private(set) annotation for public fields

OptimoveSDK/Sources/Classes/PreferenceCenter/PreferenceCenterModels.swift
```diff
@@ -0,0 +1,91 @@
...
+        channelSubscription = subscribedChannels
+    }
+
+    public required init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.topicId = try container.decode(String.self, forKey: .topicId)
+        self.topicName = try container.decode(String.self, forKey: .topicName)
+        self.topicDescription = try container.decode(String.self, forKey: .topicDescription)
+        self.channelSubscription = try container.decode([Channel].self, forKey: .channelSubscription)
+    }
```
eligutovsky: Hmm, something went wrong. Codable protocol does not require this implementation — it should be synthesized on build phase.

eligutovsky (commented):
Good job, Dani 🚀

OptimoveCore/Sources/Classes/NetworkClient/NetworkClient.swift
```diff
@@ -4,6 +4,23 @@ import Foundation
 
 public typealias NetworkServiceCompletion = (Result<NetworkResponse<Data?>, NetworkError>) -> Void
 
+extension NetworkClient {
```
vvoicehovics: there are major discrepancies VS android:

1. tenantId, region and brandGroupId are taken from base64 encoded string, not from storage set by other features which may not exist
2. this base64 encoded string is set onto config during initialisation, not passed into API calls
3. calling enablePreferenceCenter on config enables the PC feature
4. here customer Id is not checked for being null or being a visitor id
5. in android set/get preferences gives you back a result enum
6. API sits not in Optimove, but in OptimovePreferenceCenter, similar to OptimoveInApp
7. here API calls can throw / break execution and doesnt log errors VS in android we only log errors and give you respective Result into your callback
8. get/set API receives a callback as an argument VS async/await. Async/await are probably preferable in swift, but i think consistency wins. 
9. in android fetching happens on a bg thread, callbacks run on main thread. Here fetching happens on a random thread (one that is available?) it continues on the thread from which it was called

Some stuff is very debatable here, but I think we should update most (all?) for consistency

vvoicehovics: comment on consistency. on slack i was posting

```
 // 0. initialisation
    //    Optimove.initialize(this, builder)
    //            ...
    //            .enablePreferenceCenter(brandGroupId)
    //            .build();
    // 1. API
     OptimovePreferenceCenter.getInstance().getPreferencesAsync(PreferencesHandler);
     OptimovePreferenceCenter.getInstance().setCustomerPreferencesAsync(PreferencesHandler, List<PreferenceUpdate> updates);
    //    public interface PreferencesGetHandler {
    //     void run(@Nullable Preferences preferences);
    //    }
    //    public interface PreferencesSetHandler {
    //      void run(Boolean result);
    //    }
    // 3. models matching interfaces in web sdk
    // 4. what happens if no customer id? getPreferencesAsync, setCustomerPreferencesAsync immediately return, log warning, dont crash
```

which has for example
`OptimovePreferenceCenter.getInstance().setCustomerPreferencesAsync(PreferencesHandler, List<PreferenceUpdate> updates);`

vs

here we have 
`func setPreferences(for customerId: String, brandGroupId: String, _ preferenceUpdates: [PreferenceUpdate]) async throws -> Preferences`

the differences being:
1. it shouldnt take customer id
2. it shouldnt take brand group it
3. it should take callback
4. it should be called on OptimovePreferenceCenter
5. it's called differently

The point i'm trying to make is not that the interface i was proposing makes most sense (it may be total garbage), but that we should sync on the desired stuff, argue if it doesnt make --> produce smth more consistent from start --> win 🎱

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -13,12 +13,16 @@ public struct Feature: OptionSet, @unchecked Sendable, CustomStringConvertible {
...
+    public static let preferenceCenter = Feature(rawValue: 1 << 4)
     static let delayedConfiguration = Feature(rawValue: 1 << 3)
 
     public init(rawValue: Int) {
         self.rawValue = rawValue
     }
 
+
+    //Question: Is this the way to go? I believe it's what Android sdk does
     public var description: String {
```
eligutovsky: What are you asking about?

If about the `description` method, then it is a required method from the `CustomStringConvertible` protocol. It creates a string representation of any object

daniela-opti: Ahh, sorry for being unclear. I was referring to the general approach on features

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -61,27 +62,41 @@ typealias Logger = OptimoveCore.Logger
                     throw GuardError.custom("Failed on OptimobileSDK initialization. Reason: \(error.localizedDescription)")
                 }
             }
+        }   
+
+        if config.isPreferenceCenterConfigured() {
+            shared.container.resolve { serviceLocator in
+                do {
+                    try OptimovePreferenceCenter.initialize(config: config, storage: serviceLocator.storage(), networkClient: NetworkClientImpl())
```
vvoicehovics: this uses storage to get customerID. but when is customerID set / which features are required for it to be set?

i think optimove is needed. e.g. check this condition in android

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -61,27 +62,41 @@ typealias Logger = OptimoveCore.Logger
...
+                    throw GuardError.custom("Failed on PreferenceCenter initialization. Reason: \(error.localizedDescription)")
+                }
+            }
         }
     }
 
     /// Set the credentials for the Optimove server. Intent to use as a step for the delayed initialization.
-    public static func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?) {
+    @available(iOS 13.0, *)
+    public static func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?, preferenceCenterCredentials: String?) {
```
vvoicehovics: did we just casually break apps for all existing clients using delayed init? this requires 3rd argument (or i dont know swift 🤔 ?). to not require it need to pass default = nil but better to create overload like in android

https://stackoverflow.com/questions/37305951/default-optional-parameter-in-swift-function

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -47,6 +53,10 @@ public struct OptimoveConfig {
     func isOptimobileConfigured() -> Bool {
         return features.contains(.optimobile)
     }
+    
+    func isPreferenceCenterConfigured() -> Bool {
+        return features.contains(.optimobile)
```
vvoicehovics: optimobile?

daniela-opti: oopsie, copypasteerror 😅

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -105,10 +115,13 @@ open class OptimoveConfigBuilder: NSObject {
     private var _runtimeInfo: [String: AnyObject]?
     private var _sdkInfo: [String: AnyObject]?
     private var _isRelease: Bool?
+    private var environment: String?
+    private var tenantId: Int?
+    private var brandGroupId: String?
 
-    public convenience init(optimoveCredentials: String?, optimobileCredentials: String?) {
+    public convenience init(optimoveCredentials: String?, optimobileCredentials: String?, preferenceCenterCredentials: String? = nil) {
```
vvoicehovics: zzz, where is `enablePreferenceCenter()` ? PC credentials are deliberately not included in builder constructor arguments because this is a "niche" feature compared to other 2 and didnt want to clutter everyones initialisation with it. Argument list forever growing with an addition of every new feature is i think less neat than calling `enable` on the builder

same reason for having an overload when doing delayed init

Well, the reasoning can be dubious or not as usual, but that's what we have in android and slack

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -427,3 +472,50 @@ struct OptimobileArguments: Decodable {
...
+            case .emptyBase64:
+                return "The base64 string is empty"
+            case let .failedDecodingBase64(string):
+                return "Failed on decoding base64 the value \(string)"
+            }
+        }
+    }
+
+    let version: String
+    let environment: String //TODO: Change this to enum
```
vvoicehovics: in android it doesnt even do any mapping -- it just uses in url whatever is in the string. this is perhaps a little more flexible -- we dont have hardcoded PC regions in the sdk and we can e.g. easily update dev-pb --> dev

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/PreferenceUpdateRequest.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+public struct PreferenceUpdateRequest: Codable {
```
vvoicehovics: PreferenceUpdate in android

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/PreferenceUpdateRequest.swift
```diff
@@ -0,0 +1,11 @@
...
+    public let topicId: String
+    public let channelSubscription: [Channel]
+
+    public init(topicId: String, channelSubscription: [Channel]) {
+        self.topicId = topicId
+        self.channelSubscription = channelSubscription
```
vvoicehovics: and it's also not "channelSubscription" here . check web sdk interfaces

```
export interface Preferences {
    configuredChannels: Channel[];
    customerPreferences: Topic[];
}

type TopicId = string;

interface Topic {
    id: TopicId;
    name: string;
    description: string;
    subscribedChannels: Channel[];
}

export interface PreferenceUpdate {
    topicId: TopicId;
    subscribedChannels: Channel[];
}
```

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/Preferences.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+public struct Preferences: Codable {
+    public struct Topic: Codable {
+        public let topicId: String
```
vvoicehovics: and same here. doesnt match interfaces in web sdk.

can we check everywhere and remap what needs to be remapped

OptimoveSDK/Sources/Classes/Services/ServiceLocator.swift
```diff
@@ -107,6 +107,10 @@ final class ServiceLocator {
         return _deviceStateObserver
     }
 
+//    @available(iOS 13.0, *)
+//    func preferenceCenter() -> PreferenceCenter {
+//        return PreferenceCenter(storage: storage(), networkClient: NetworkClientImpl())
+//    }
```
vvoicehovics: 🔪 or 🌞

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,198 @@
...
+            case .alreadyInitialized:
+                return "The PreferenceCenterSDK has already been initialized."
+            case .configurationIsMissing:
+                return "Preference center config is missing."
+            }
+        }
+    }
+
+    private static var instance: OptimovePreferenceCenter?
+    private var config: PreferenceCenterConfig?
```
vvoicehovics: 1 comment i had in android recently, is that if delayed initialisation is updaing OptimoveConfig, we need to make sure PreferencCenterConfig gets updated too. to not breed copies/references, can store OptimoveConfig here

or even better not store it at all accessing as Optimove.getConfigSomehow().getPreferenceCenterConfig() when needed

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,198 @@
...
+        self.config = config
+        self.networkClient = networkClient
+        self.storage = storage
+
+        Logger.debug("Preference center SDK was initialized with \(config)")
+    }
+
+    public func getPreferencesAsync(completion: @escaping PreferencesGetHandler) {
+        guard let customerId = try? storage?.getCustomerID(), let visitorId = try? storage?.getVisitorID() else {
+            completion(.errorUserNotSet, nil)
```
vvoicehovics: Logger.warn("Customer ID is not set") also here?

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,198 @@
...
+            } catch {
+                DispatchQueue.main.async {
+                    completion(.error, nil)
+                }
+            }
+        }
+    }
+
+    private func fetchPreferences(customerId: String) async throws -> Preferences {
+        guard config != nil else { throw Error.configurationIsMissing }
```
vvoicehovics: we want every error to be logged. can double check everywhere?

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/PreferenceUpdateRequest.swift
daniela-opti: can't we update Android for this one? `PreferenceUpdate` on its own doesn't sound super clear. It took Eli a few extra seconds to understand what it is, which is a sign that it could use an improvement :p

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -178,6 +195,21 @@ open class OptimoveConfigBuilder: NSObject {
             Logger.error(error.localizedDescription)
         }
 
+        return self
+    } 
+
+    @discardableResult public func setCredentials(
```
vvoicehovics: does this need to be public? customers should not be doing builder.setCredentials() -- they either proovide them immediately or Optimove.setCredentials 

(maybe it's some limitation we had / ?)

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -299,18 +343,53 @@ open class OptimoveConfigBuilder: NSObject {
             }
             Logger.info("\(OptimobileConfig.self) building skipped.")
             return nil
-        }()
+        }()    
+
+        let preferenceCenterConfig: PreferenceCenterConfig? = getPreferenceCenterConfig()
```
vvoicehovics: if someone just doesnt want PC, are we going to aways log error?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -299,18 +343,53 @@ open class OptimoveConfigBuilder: NSObject {
...
+                let args = try PreferenceCenterArguments(base64: credentials)
+
+                return PreferenceCenterConfig(
+                    region: args.region,
+                    tenantId: args.tenantId,
+                    brandGroupId: args.brandGroupId
+                )
+
+            } catch {
+                //I see Android throws, but here we don't (I don't see throw in the build methods at all), should I be adding it?
```
vvoicehovics: to me it makes more sense throwing -- like if they provide garbage string, better to throw than silently not init PC

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/PreferenceUpdateRequest.swift
vvoicehovics: 1. also web sdk has PreferenceUpdate
2. to me 'request' is smth going to Web unlike this object. Same way we could be attaching "Request" ~ to every argument, because, well, they passed some data into function that updates stuff, so, "requested". Or what other meaning does 'Request' add here?
3. while i dont like `PreferenceUpdateRequest`, if it's not clear we could rename it to some kinda `TopicSubscriptionsUpdate` or just `TopicSubscriptions` (but TopicSubscriptions[] feels odd) or a better name. But then it's on you to i) convince everyone other name is better ii) update all sdks ----> and idk why do we need to go there 😓

daniela-opti: well argued 😸  I will leave it as PreferenceUpdate, and if we really don't like it in the long run, I'll consider myself responsible for updating all sdks

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
daniela-opti: Hmm, I'm not sure; seems to work when I remove the `public` keyword. @eligutovsky would that break something I am not aware of?

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
daniela-opti: @eligutovsky thoughts on this? Will most likely add the `throws` and handle, but wondering if there was a reason behind this? (build method in `OptimoveConfig` not throwing)

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,191 @@
...
+
+    public static func getInstance() throws -> OptimovePreferenceCenter {
+        guard let instance = instance else {
+            throw Error.configurationIsMissing
+        }
+        return instance
+    }
+
+    static func initialize(storage: OptimoveStorage, networkClient: NetworkClient) throws {
+        guard instance == nil else {
```
vvoicehovics, line 55: 1. we do delayed init, we call Optimove.initialize, which calls OptimovePreferenceCenter.initialize
2. then we do setCredentials, which calls initialize(with: config), which calls OptimovePreferenceCenter.initialize
3. it throws an error here, which we log
----> every delayed init of PC logs an error

moral
1. this happens because android and ios work slightly differently. On android with delayed init we dont call Optimove.initialize again, on ios we do. With such little differences it becomes harder to verify 2 logics -- to me having them as close as possible is ideal, even when it sometimes deviates from language best practices (but ofc still within language acceptable practices :) ) 
2. we should not only make working code, but also care about best possible API / error logging -- customer experience -- which is a direct source of inquiries. This involves deciding what we should log and how, actually checking that correct errors logged, having sensible error messages etc. Just feels logging is not getting enough attention over and over again, here and in android sdk and in general

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -178,6 +195,21 @@ open class OptimoveConfigBuilder: NSObject {
...
 
+        return self
+    } 
+
+    @discardableResult func setCredentials(
+        optimoveCredentials: String?,
+        optimobileCredentials: String?,
+        preferenceCenterCredentials: String?
+    ) -> OptimoveConfigBuilder {
+        setCredentials(optimoveCredentials: optimoveCredentials, optimobileCredentials: optimobileCredentials)
```
vvoicehovics: this one checks that optimove or optimobile credentials are not nil. so, you cannot use it as setCredentials(nil, nil, credentials)

we need to make sure all these permutations of nullability possibilities make some sense

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -153,10 +168,11 @@ open class OptimoveConfigBuilder: NSObject {
         super.init()
     }
 
-    @discardableResult public func setCredentials(optimoveCredentials: String?, optimobileCredentials: String?) -> OptimoveConfigBuilder {
```
vvoicehovics, line 156: technically this is breaking public contract now :) but assuming nobody should have been using it + it's not documented + only 1-2 clients use delayed init + on android we dont have it + if they did use it would break app and they would spot i think (well, the last arg is not so strong, but the point is in total i think it's very very unlikely)

1. proceed with the change: having 2 ways to set credentials on builder is not lean / confusing / raises q "which one should i use"
```
  let config = OptimoveConfigBuilder(optimoveCredentials: "", optimobileCredentials: "")
            .setCredentials(optimoveCredentials: "", optimobileCredentials: "")
```
2. dont have it as major release to not scare all customers from updating

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -178,6 +195,21 @@ open class OptimoveConfigBuilder: NSObject {
...
+    @discardableResult func setCredentials(
+        optimoveCredentials: String?,
+        optimobileCredentials: String?,
+        preferenceCenterCredentials: String?
+    ) -> OptimoveConfigBuilder {
+        setCredentials(optimoveCredentials: optimoveCredentials, optimobileCredentials: optimobileCredentials)
+        
+        if let preferenceCenterCredentials = preferenceCenterCredentials, !preferenceCenterCredentials.isEmpty {
+            self.preferenceCenterCredentials = preferenceCenterCredentials
+            features.insert(.preferenceCenter)
```
vvoicehovics: the point of features was separating which credentials they have now VS which features they want in general. so, when they init initially, they say we want features A, B, C. Then they can provide credentials for them later

hence, 
- no need to add a feature when setting credentials
- if set credentials for feature that was not requested -- error

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -299,18 +339,48 @@ open class OptimoveConfigBuilder: NSObject {
             }
             Logger.info("\(OptimobileConfig.self) building skipped.")
             return nil
-        }()
+        }()    
+
+        let preferenceCenterConfig: PreferenceCenterConfig? = features.contains(.preferenceCenter) ? getPreferenceCenterConfig() : nil
```
vvoicehovics: also this line i think doesnt work (?). 

if we use delayed initialisation containing PC in feature set, we would try to get config here. But because it's a delayed init we dont have credentials -- it would log error. But it's not an error, it's a legit delayed init

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
vvoicehovics: i think the reason was smth along "we want to do everything not to crash the app" . i guess both ways possible

in android and previously in all mobile sdks the idea was that if they give bad credentials -- our sdk cannot function correctly, it could create more inquiries, be more obscure to customers. Since they provide credentials during initialization, if they do any app testing at all -- it would be very hard for them to miss the crash / release app with bad credentials.

After delayed init was added, guess it is more possible if credential strings can come from somewhere dynamically. If this actually happened and someone released app which could load bad PC credentials -- would we like to crash the app for them? Probably not.

so, mind change :) let's keep ios sdk consistent within itself and never crash if they give bad strings . Android will check, may also adjust for PC only. Or may do a field test -- which one works better :)

vvoicehovics: but if this actually happened in testing -- crashing would be better. 

mb behaviour could even be 

prod -- dont crash, dev -- crash
delayed -- dont crash, normal init -- crash

So, the answer kinda depends on how well they test their apps / how much more time we want to spend on crash dont crash.

since i think not much, let's not crash on ios

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
daniela-opti: ahh! This is an important piece of information I wasn't aware of! Thanks for sharing 🙌

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
daniela-opti: Agree with this, but not sure what to do to resolve this.

Are you suggesting removing the `Logger.error("Could not find preference center credentials")` below and just returning `nil`? That way we just won't proceed with pref center initialisation and won't notify anyone? Or another solution? @vvoicehovics

OptimoveSDK/Sources/Classes/PreferenceCenter/Models/Preferences.swift
```diff
@@ -0,0 +1,37 @@
...
+
+        enum CodingKeys: String, CodingKey {
+            case id = "topicId"
+            case name = "topicName"
+            case description = "topicDescription"
+            case subscribedChannels = "channelSubscription"
+        }
+    }
+
+    public let topics: [Topic]
```
vvoicehovics: comon, Daniela, how hard is it to take this interface

```
export interface Preferences {
    configuredChannels: Channel[];
    customerPreferences: Topic[];
}
```

and have it _exactly the same_ in ios ?

The reason for it being called `customerPreferences` and not `topics` is that it contains subscriptions of this customer to topics, not only data about the topics

daniela-opti: I don't appreciate the tone. It's literally a one click change and there were more important things I was focusing on. I have forgotten to update it, yes, but there is no need to be derogatory :)

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
daniela-opti: I've added this, so nothing gets logged during delayed init

```
   let preferenceCenterConfig: PreferenceCenterConfig? = {
            if features.contains(.preferenceCenter),
               preferenceCenterCredentials != nil
            {
```

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
daniela-opti: Wait, but do we not want at least one? Or do we want to be able to call with `nil, nil, prefCenterCredentials`?

daniela-opti: Okay, that is possible after the latest change

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
vvoicehovics: can check what other features do, e.g. optimobile

```
    static func initialize(config optimoveConfig: OptimoveConfig, initialVisitorId: String, initialUserId: String?) throws {
        if instance !== nil, optimoveConfig.features.contains(.delayedConfiguration) {
            try completeDelayedConfiguration(config: optimoveConfig.optimobileConfig!)
            return
        }
```

so
a) if it's already initialized and it is a delayed init -- skip
b) even simpler, if it's already initialized -- just dont do anything. credentials dont affect initialisation

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        instance = OptimovePreferenceCenter(storage: storage, networkClient: networkClient)
+    }
+
+    private init(storage: OptimoveStorage, networkClient: NetworkClient) {
+        self.networkClient = networkClient
+        self.storage = storage
+
+        Logger.debug("Preference center SDK was initialized")
```
vvoicehovics: need this one?

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,192 @@
...
+        else {
+            Logger.warn("Customer ID is not set")
+            completion(.errorUserNotSet, nil)
+            return
+        }
+
+        Task {
+            do {
+                let request = createGetPreferencesRequest(for: customerId, with: config)
+                let data = try await networkClient?.performAsync(request) ?? Data()
```
vvoicehovics: in which cases ?? Data() happens? what happens next, when decoder tries to get Preferences from it?

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,192 @@
...
+                    HTTPHeader(field: .accept, value: .textplain),
+                    HTTPHeader(field: .tenantId, value: .tenantId(id: tenantId))
+                ],
+                queryItems: [
+                    URLQueryItem(name: "customerId", value: customerId),
+                    URLQueryItem(name: "brandGroupId", value: brandGroupId)
+                ]
+            )
+        } catch {
+            fatalError("Configuration is missing. This should never happen if the SDK is initialized properly.")
```
vvoicehovics, line 112: does this one crash the app? If we go consistently about "PC never crashes the app" -- shall just log

vvoicehovics: this one insta-crashes the app. If we go consistently about PC cannot crash it ever -- should just log error / call handler with .error

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,197 @@
...
+    public static func getInstance() throws -> OptimovePreferenceCenter {
+        guard let instance = instance else {
+            throw Error.notInitialized
+        }
+        return instance
+    }
+
+    static func initialize(storage: OptimoveStorage, networkClient: NetworkClient) throws {
+        if let config = Optimove.getConfig() {
+            if instance !== nil && config.features.contains(.delayedConfiguration) {
```
daniela-opti: this actually might be redundant because we have this check in `OptimoveConfig`

```swift
    let preferenceCenterConfig: PreferenceCenterConfig? = {
            guard features.contains(.preferenceCenter) else {
                if preferenceCenterCredentials != nil {
                    Logger.error("Preference center credentials provided, but the feature was not requested.")
                }
                return nil
            }
            
              // referring to this
            if preferenceCenterCredentials == nil {
                if !features.contains(.delayedConfiguration) {
                    Logger.error("Preference center could not be initialized due to missing credentials.")
                }
                return nil
            }

            return getPreferenceCenterConfig(from: preferenceCenterCredentials)
        }()

```

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,197 @@
...
+                    HTTPHeader(field: .tenantId, value: .tenantId(id: tenantId))
+                ],
+                queryItems: [
+                    URLQueryItem(name: "customerId", value: customerId),
+                    URLQueryItem(name: "brandGroupId", value: brandGroupId)
+                ]
+            )
+    }
+
+    private func getConfigValues(from config: PreferenceCenterConfig) throws -> (region: String, brandGroupId: String, tenantId: String) {
```
vvoicehovics: does this ever throw? PreferenceCenterConfig can only be of right shape or nil?

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,197 @@
...
+        let region = config.region
+        let brandGroupId = config.brandGroupId
+        let tenantId = config.tenantId.description
+        return (region, brandGroupId, tenantId)
+    }
+
+
+    public func setPreferencesAsync(updates: [OptimovePC.PreferenceUpdate], completion: @escaping PreferencesSetHandler) {
```
vvoicehovics: one from android is `public void setCustomerPreferencesAsync(@NonNull PreferencesSetHandler preferencesSetHandler, List<PreferenceUpdate> updates) {`

1. arg order
2. diff name

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -0,0 +1,197 @@
...
+        Task {
+            do {
+                let request = try createSetPreferencesRequest(for: customerId, with: config, updates: updates)
+                _ = try await networkClient?.performAsync(request)
+                
+                DispatchQueue.main.async {
+                    completion(.success)
+                }
+            } catch {
+                DispatchQueue.main.async {
```
vvoicehovics, line 193: here we dont log any error

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
daniela-opti: @vvoicehovics Android has
- `getPreferencesAsync`
- `setCustomerPreferencesAsync`

Should we update it to `getCustomerPreferencesAsync`?

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
vvoicehovics: it might be redundant because of this

```
    func isOptimobileConfigured() -> Bool {
        return features.contains(.optimobile)
    }
    
    func isPreferenceCenterConfigured() -> Bool {
        return features.contains(.preferenceCenter) && preferenceCenterConfig != nil
    }

```

because of ^ it's not going to be initialised. But ^ is not entirely correct -- `isXConfigured` checks if we want it in general, and if we do  for other modules it can start doing work before credentials given. 

So, a module should be initialised when it is configured, and it is configured when it is among features. And then when credentials given it would be initialised again

lol, this is confusing :) but i think if we try to keep modules behaving more similar, will be slightly simpler to reason about

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
vvoicehovics: suppose no, it was intentional -- back to that Preferences ambiguity

- `getPreferences` gives you customer preferences, but also some PC configuration. Later it may give you more random stuff, we just get some ill-defined "Preferences"
- `setCustomerPreferences` highlights that we are setting only preferences for the customer, nothing else. Also in Preferences we have key `customerPreferences` -- they kinda match

web and android do same. Reasoning dubious as usual :) but let's keep consistent with rest

OptimoveSDK/Sources/Classes/Optimove.swift
```diff
@@ -61,27 +62,70 @@ typealias Logger = OptimoveCore.Logger
                     throw GuardError.custom("Failed on OptimobileSDK initialization. Reason: \(error.localizedDescription)")
                 }
             }
+        }   
+
+        if config.isPreferenceCenterConfigured() {
+            guard config.isOptimoveConfigured() else {
+                Logger.error("Preference center requires optimove credentials")
```
vvoicehovics: this log is not 100% correct -- not "credentials" but "feature enabled" or both -- as for delayed init we dont require credentials

OptimoveSDK/Sources/Classes/PreferenceCenter/OptimovePreferenceCenter.swift
```diff
@@ -86,14 +86,29 @@ public class OptimovePreferenceCenter {
...
             do {
                 let request = try createGetPreferencesRequest(for: customerId, with: config)
-                guard let data = try await networkClient?.performAsync(request) else {
-                    throw NetworkError.noData
-                }
-                let preferences = try JSONDecoder().decode(OptimovePC.Preferences.self, from: data)
 
-                DispatchQueue.main.async {
-                    completion(.success, preferences)
+                networkClient?.perform(request) { [self] result in
```
vvoicehovics, line 90: if we remove client now, we dont need available(13) all over the place? only in PC

daniela-opti: Resolved

vvoicehovics (approved):

PR #114, opened 2024-11-01
AB#232967 Fix immediate event sync edge case

cgwyllie:
### Description of Changes

In some cases, depending on combinations of SDK method calls, the same event may be synced to the backend more than once.

Make event sync more robust against this to improve performance and reduce redundancy.

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [x] `OptimoveCore.podspec`
- [x] `OptimoveSDK.podspec`

- [x] `README.md`
- [x] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [x] Init SDK with all credentials
- [x] Track events
- [x] Associate customer (verify both backends)
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:

### DIFF
--- a/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
+++ b/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
@@ -164,7 +164,7 @@ final class AnalyticsHelper {
 
     private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
         self.serialQueue.async {
-            self.syncEventsImpl(context: self.analyticsContext, onSyncComplete)
+            self.syncEventsImpl(context: context, onSyncComplete)
         }
     }
 
--- a/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
+++ b/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
@@ -18,12 +18,14 @@ final class AnalyticsHelper {
     private var analyticsContext: NSManagedObjectContext?
     private var migrationAnalyticsContext: NSManagedObjectContext?
     private var finishedInitializationToken: NSObjectProtocol?
+    private let syncQueue : DispatchQueue
 
     // MARK: Initialization
 
     init(httpClient: KSHttpClient) {
         analyticsContext = nil
         migrationAnalyticsContext = nil
+        syncQueue = DispatchQueue(label: "com.optimove.optimobile.sync-q", target: .global(qos: .utility))
 
         eventsHttpClient = httpClient
 
@@ -149,9 +151,7 @@ final class AnalyticsHelper {
                 try context.save()
 
                 if immediateFlush {
-                    DispatchQueue.global().async {
-                        self.syncEvents(context: self.analyticsContext, onSyncComplete)
-                    }
+                    self.syncEvents(context: self.analyticsContext, onSyncComplete)
                 }
             } catch {
                 print("Failed to record event")
@@ -163,6 +163,16 @@ final class AnalyticsHelper {
     }
 
     private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
+        // Adding a small delay here can result in better grouping of sync calls -- should we do it?
+        self.syncQueue.asyncAfter(deadline: .now() + 1) {
+            let uuid = UUID().uuidString
+            print("begin \(uuid)")
+            self.syncEventsImpl(context: self.analyticsContext, onSyncComplete)
+            print("end \(uuid)")
+        }
+    }
+
+    private func syncEventsImpl(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
         context?.performAndWait {
             let results = fetchEventsBatch(context)
 
@@ -224,23 +234,44 @@ final class AnalyticsHelper {
             eventIds.append(event.objectID)
         }
 
+        // TODO clean up logs
+        print("Syncing \(events.map({$0.uuid})) events")
+
         let path = "/v1/app-installs/\(OptimobileHelper.installId)/events"
 
+        var err : Error? = nil
+        let networkBarrier = DispatchSemaphore(value: 1)
+
         eventsHttpClient.sendRequest(.POST, toPath: path, data: data, onSuccess: { _, _ in
-            if let err = self.pruneEventsBatch(context, eventIds) {
-                print("Failed to prune events batch: " + err.localizedDescription)
-                onSyncComplete?(err)
-                return
-            }
-            self.syncEvents(context: context, onSyncComplete)
+            networkBarrier.signal()
         }) { _, error, _ in
             print("Failed to send events")
-            onSyncComplete?(error)
+            err = error
+            networkBarrier.signal()
         }
+
+        print("waiting on network")
+        networkBarrier.wait()
+        print("continuing...")
+
+        if err != nil {
+            onSyncComplete?(err)
+            return
+        }
+
+        if let err = self.pruneEventsBatch(context, eventIds) {
+            print("Failed to prune events batch: " + err.localizedDescription)
+            onSyncComplete?(err)
+            return
+        }
+
+        self.syncEventsImpl(context: context, onSyncComplete)
     }
 
     private func pruneEventsBatch(_ context: NSManagedObjectContext?, _ eventIds: [NSManagedObjectID]) -> Error? {
         var err: Error?
+
+        print("Pruning \(eventIds.count)")
 
         context?.performAndWait {
             let request = NSBatchDeleteRequest(objectIDs: eventIds)
--- a/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
+++ b/OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
@@ -163,12 +163,8 @@ final class AnalyticsHelper {
     }
 
     private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
-        // Adding a small delay here can result in better grouping of sync calls -- should we do it?
-        self.syncQueue.asyncAfter(deadline: .now() + 1) {
-            let uuid = UUID().uuidString
-            print("begin \(uuid)")
+        self.syncQueue.async {
             self.syncEventsImpl(context: self.analyticsContext, onSyncComplete)
-            print("end \(uuid)")
         }
     }
 
@@ -234,25 +230,19 @@ final class AnalyticsHelper {
             eventIds.append(event.objectID)
         }
 
-        // TODO clean up logs
-        print("Syncing \(events.map({$0.uuid})) events")
-
         let path = "/v1/app-installs/\(OptimobileHelper.installId)/events"
 
         var err : Error? = nil
-        let networkBarrier = DispatchSemaphore(value: 1)
+        let networkBarrier = DispatchSemaphore(value: 0)
 
         eventsHttpClient.sendRequest(.POST, toPath: path, data: data, onSuccess: { _, _ in
             networkBarrier.signal()
         }) { _, error, _ in
-            print("Failed to send events")
             err = error
             networkBarrier.signal()
         }
 
-        print("waiting on network")
         networkBarrier.wait()
-        print("continuing...")
 
         if err != nil {
             onSyncComplete?(err)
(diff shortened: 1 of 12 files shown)


OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+    override func tearDown()
+    {
+        analyticsHelper = nil
+        mockHttpClient = nil
+        super.tearDown()
+    }
+    
+    func test_number_of_sent_events_same_as_tracked() {
+        let numberOfEvents = 8;
```
daniela-opti: I know this is super random and unnecessary, but there are a bunch of semicolons we might not need hahah (3 in this file)

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        mockHttpClient.forward = { method, data in
+            if let data = data as? [[String: Any?]], data.count == numberOfEvents {
+                numberOfEventsExpectation.fulfill()
+            }
+        }
+        
+        self.analyticsHelper.trackEvent(eventType: "immediate_event", properties: nil, immediateFlush: false)
+        self.analyticsHelper.trackEvent(eventType: "regular_event", properties: nil, immediateFlush: true)
+        
+        waitForExpectations(timeout: defaultTimeout, handler: nil)
```
daniela-opti: To clarify my own understanding - where does defaultTimeout come from?

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        let numberOfEventsExpectation = expectation(description: "Number of events wasnt \(numberOfEvents)")
+        
+        mockHttpClient.forward = { method, data in
+            if let data = data as? [[String: Any?]], data.count == numberOfEvents {
+                numberOfEventsExpectation.fulfill()
+            }
+        }
+        
+        for i in 1...numberOfEvents {
+            self.analyticsHelper.trackEvent(eventType: "immediate_event\(i)", properties: nil, immediateFlush: true)
```
daniela-opti: do we need `self` in `analyticsHelper.trackEvent`? I know it's usually optional, but wondering if there is a reason here

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    
+}
+
+class MockKSHttpClient: KSHttpClient {
+    var forward: ((_ method: OptimoveSDK.KSHttpMethod, _ data: Any?) -> Void)?
+    
+    func sendRequest(_ method: OptimoveSDK.KSHttpMethod, toPath path: String, data: Any?, onSuccess: @escaping OptimoveSDK.KSHttpSuccessBlock, onFailure: @escaping OptimoveSDK.KSHttpFailureBlock) {
+        forward?(method, data)
+        onSuccess(nil, nil)
+    }
```
daniela-opti, line 130: I noticed we're not calling `onFailure` anywhere. Should we maybe add some cases where we trigger `onFailure` to test how AnalyticsHelper handles errors? Or is it not worth doing? You are more aware of the context

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -226,17 +232,30 @@ final class AnalyticsHelper {
...
-            self.syncEvents(context: context, onSyncComplete)
+            networkBarrier.signal()
         }) { _, error, _ in
-            print("Failed to send events")
-            onSyncComplete?(error)
+            err = error
+            networkBarrier.signal()
+        }
+
+        networkBarrier.wait()
```
daniela-opti, line 246: Could using `DispatchSemaphore` here might lead to deadlocks? E.g. if the `sendRequest` callbacks are executed on the same `serialQueue` that's being blocked by `networkBarrier.wait()`?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -163,6 +163,12 @@ final class AnalyticsHelper {
     }
 
     private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
+        self.serialQueue.async {
+            self.syncEventsImpl(context: self.analyticsContext, onSyncComplete)
+        }
+    }
```
daniela-opti, line 169: In `syncEvents`,  we're passing `self.analyticsContext` to `syncEventsImpl` instead of the context parameter received by the method. Is there a specific reason for this? 
Would passing the context parameter to `syncEventsImpl` be an option?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -226,17 +232,30 @@ final class AnalyticsHelper {
...
-                onSyncComplete?(err)
-                return
-            }
-            self.syncEvents(context: context, onSyncComplete)
+            networkBarrier.signal()
         }) { _, error, _ in
-            print("Failed to send events")
-            onSyncComplete?(error)
+            err = error
+            networkBarrier.signal()
```
daniela-opti, line 243: I noticed we're assigning the error to `err` in the failure callback but not logging it. Should we add a print statement or some form of logging here to help with debugging?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -18,12 +18,14 @@ final class AnalyticsHelper {
...
     private var migrationAnalyticsContext: NSManagedObjectContext?
     private var finishedInitializationToken: NSObjectProtocol?
+    private let serialQueue : DispatchQueue
 
     // MARK: Initialization
 
     init(httpClient: KSHttpClient) {
         analyticsContext = nil
         migrationAnalyticsContext = nil
+        serialQueue = DispatchQueue(label: "com.optimove.optimobile.serial-q", target: .global(qos: .utility))
```
daniela-opti, line 28: For my own understanding - what's the reason for setting the target of `serialQueue` to `.global(qos: .utility)`? Do we need to dispatch our serial queue onto the global utility queue?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
```diff
@@ -163,6 +163,12 @@ final class AnalyticsHelper {
     }
 
     private func syncEvents(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
+        self.serialQueue.async {
+            self.syncEventsImpl(context: self.analyticsContext, onSyncComplete)
+        }
+    }
+
+    private func syncEventsImpl(context: NSManagedObjectContext?, _ onSyncComplete: SyncCompletedBlock? = nil) {
         context?.performAndWait {
```
daniela-opti, line 172: Another thing to improve my general understanding/knowledge - since we're using `performAndWait`, could this cause threading issues or deadlocks? Would `perform` work?

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
k-antipochkin: :sweat_smile: fixed

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
k-antipochkin: It a variable in OptimoveTestCase

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
k-antipochkin: Removed

OptimoveSDK/Tests/Sources/Optimobile/AnalyticsHelperTests.swift
k-antipochkin: Added a failure test case

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
k-antipochkin: It would happen if the network request completion handler would run on the serial queue we are waiting the semaphore on. The completion / error handlers arent run on our private serial queue

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
k-antipochkin: Trying to avoid any new prints. (Used to have a logging mechanism that allowed the log level control)

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
k-antipochkin: Although wasn't changed, it is very important to have  `performAndWait` it is done on the current thread (serial queue), guaranteeing that the block will finish before next one starts

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
k-antipochkin: utility qos means low priority. Do you think we need some other qos?

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
k-antipochkin: Changed

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
daniela-opti: Got it 👌 thanks for clarifying

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
daniela-opti: Gotcha, `.utility` keeps it low priority. I'm not exactly sure of what the benefits would be, I can't offer a lot of experience here. I'd say I trust your solution - you have a better understanding of the problem and the bigger picture

daniela-opti (approved):
Looks good to me, so as long as tests are working in the pipeline, it should be good to go

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
cgwyllie: Funnily I was just coming to post about QoS and review: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/PrioritizeWorkAtTheTaskLevel.html

I think utility is probably acceptable here, we don't want it as low priority as background perhaps, but it's definitely not as important as user-initiated UI interactions.

RE your Q on the use of a global queue as the target, this just allows the DispatchQueue to re-use existing global thread pools to execute work rather than creating its own so it's more efficient. The DispatchQueue still ensures there is a 'serial' execution of tasks in the queue, but it may use different threads form the global pool.

Hope that makes sense, @k-antipochkin correct me if I'm mistaken!

OptimoveSDK/Sources/Classes/Optimobile/AnalyticsHelper.swift
cgwyllie: print() is only visible on the debugger I think, so it is only useful for developer feedback, so I'd vote to keep prints for SDK development purposes / app developer debug builds. WDYT?

Shared/Sources/Extension/Bundle.swift
```diff
@@ -10,7 +10,7 @@ public extension Bundle {
             if let moduleName = Bundle(for: BundleFinder.self).bundleIdentifier,
                let testBundlePath = ProcessInfo.processInfo.environment["XCTestBundlePath"]
             {
-                if let resourceBundle = Bundle(path: testBundlePath + "/Optimove_\(moduleName).bundle") {
+                if let lastModuleComponent = moduleName.components(separatedBy: ".").last, let resourceBundle = Bundle(path: testBundlePath + "/Optimove_\(lastModuleComponent).bundle") {
```
cgwyllie, line 13: Hey @k-antipochkin , small Q about what this change is for?

k-antipochkin: The tests are failing in Xcode 16+, cant find the test resource files. The module name becomes sdk.OptimoveTests.

k-antipochkin (approved):

PR #122, opened 2025-04-14
244691 - Added Embedded Messaging Ability to IOS SDK

adam-b-optimove:
[244691](https://mobius.visualstudio.com/Backstage/_workitems/edit/244691)

### Description of Changes

- Added new requests for the embedded messaging feature
- Added Unit tests for changes

Test Run:

https://optimove.testrail.io/index.php?/runs/view/327&group_by=cases:title&group_order=asc

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [ ] `OptimoveCore.podspec`
- [ ] `OptimoveSDK.podspec`

- [ ] `README.md`
- [ ] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -15,6 +15,9 @@ public struct Feature: OptionSet, @unchecked Sendable, CustomStringConvertible {
     public static let optimove = Feature(rawValue: 1 << 2)
     /// Preference center feature.
     public static let preferenceCenter = Feature(rawValue: 1 << 4)
+    /// Embedded messaing feature
```
daniel-c-opti, line 18: if I understand this right we need to use a different number other than 4 as that is in use for the preference center.

daniel-c-opti (commented):
quick comment I noticed. There's a bit of commented out code and some hard coded values in places. I will hold off on a full review until you say you've done the code cleanup

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
adam-b-optimove: updated this

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessageResponse.swift
```diff
@@ -0,0 +1,8 @@
+public typealias EmbeddedMessagesResponse = [String: EmbeddedMessagingContainer]
```
daniel-c-opti: Kind of a style question but why do we have so many models in the EmbeddedMessage.swift file and then seperate files for the other models, might be worth either having them all in one file or all seperate files

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
```diff
@@ -356,13 +389,32 @@ open class OptimoveConfigBuilder: NSObject {
 
             return getPreferenceCenterConfig(from: preferenceCenterCredentials)
         }()
+        
+        let embeddedMessagingConfig: EmbeddedMessagingConfig? = {
+//            if !features.contains(.optimove),
```
daniel-c-opti: should this code be commented out still?

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,467 @@
...
+        let brandId = config.brandId
+        let tenantId = config.tenantId.description
+        return (region, brandId, tenantId)
+    }
+    
+    
+    // MARK: - Get Messages
+    public func getMessagesAsync(
+        completion: @escaping EmbeddedMessagingGetHandler,
+        containers: [EmbeddedMessageOptions]? = nil
```
daniel-c-opti: shouldn't this be called ContainerRequestOptions according to the chat we had the other day.

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,467 @@
...
+            completion(.errorUserNotSet)
+            return
+        }
+
+        do {
+            let request = try createDeleteMessagesRequest(customerId: customerId, visitorId: visitorId, config: config, messageId: embeddedMessage.id)
+            
+            networkClient?.perform(request) { result in
+                switch result {
+                case .success(let response):
```
daniel-c-opti: github actions makes a decent point about the response here. Can we make it _ or just not have it?

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,467 @@
...
+        } catch {
+            logFailedResponse(error)
+            DispatchQueue.main.async {
+                completion(.error(.errorSendingRequest))
+            }
+        }
+    }
+    
+    // MARK: - Delete Messages
+    public func deleteMessagesAsync(completion: @escaping EmbeddedMessagingGetHandler, embeddedMessage: EmbeddedMessage, isRead: Bool = false) {
```
daniel-c-opti: I'm not 100% concerned about this but the parameter being embeddedMessage might be out of sync with what I've done in Android. If you do change it though it isn't just here but all of the public funcs here should match. up to you.

daniel-c-opti (commented):
few comments. Nice work!

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,467 @@
...
+        
+        do {
+            let request = try createGetEmbeddedMessagesRequest(
+                customerId: customerId,
+                visitorId: visitorId,
+                config: config,
+                containers: containers
+            )
+            
+            networkClient?.perform(request) { result in
```
eligutovsky, line 142: Could you verify why `networkClient` is optional here? I believe it has to be passed on init of this object

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,467 @@
...
+    private static var instance: EmbeddedMessagesService?
+    private var storage: OptimoveStorage?
+    private var networkClient: NetworkClient?
+
+    public static func getInstance() throws -> EmbeddedMessagesService {
+        guard let instance = instance else {
+            throw Error.notInitialized
+        }
+        return instance
+    }
```
eligutovsky: Maybe it is out of this PR scope, but why do we control object's lifecycle at object itself. The Service Locator (it is variation of DI) is great alternative for lifecycle management

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
eligutovsky, line 1: This class has many resposibilities. What way to decrease this complexity we can do?

eligutovsky: Could please challenge this class public visibility? Do we actually want to expose this class to customers? 

The same question to other classes and values

eligutovsky: Are you planning to add unit tests for the new code?

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: I have renamed this. good spot

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: without making it optional we get: Value of optional type '(any NetworkClient)?' must be unwrapped to refer to member 'perform' of wrapped base type 'any NetworkClient'. So iv made it optional to match the way preference center does it

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: good point, iv updated this

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessageResponse.swift
adam-b-optimove: yeah totally good point, iv moved them all to one file

OptimoveSDK/Sources/Classes/OptimoveConfig.swift
adam-b-optimove: good shout, this should have been removed

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: yeah there is likely a way to improve this, at this stage we are just following the same standards from preference center

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+    public let customerId: String
+    public let isVisitor: Bool
+    public let templateId: Int64
+    public let title: String
+    public let content: String?
+    public let media: String?
+    public let readAt: Int?
+    public let url: String?
+    public let engagementId: String
+    public let payload: [String: String]
```
cgwyllie: Hey @adam-b-optimove , I think this is correct for an inbox message, but not necessarily for an embedded message in the future.

@daniel-c-opti how did we handle payload in the Android SDK?

We should try to future-proof it for arbitrary JSON objects in payload to prevent us making breaking changes in future.

I think we had a few ideas:
- give back 'raw' payload and let developer parse it (lowest effort, not nicest DX)
- give back a 'generic' parsed type like `[AnyHashable : Any]` (similar to other parts of SDK, not most intuitive either)
- offer some `getPayload<T>()` type thing that parses payload to a Codable `T` type

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+
+public struct EmbeddedMessage: Codable {
+    public let customerId: String
+    public let isVisitor: Bool
+    public let templateId: Int64
+    public let title: String
+    public let content: String?
+    public let media: String?
+    public let readAt: Int?
```
cgwyllie: Are these types aligned across SDKs? (e.g. are we giving dates in some and timestamps in others)

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
```diff
@@ -0,0 +1,484 @@
...
+    private func createReadAtMessagesRequest(
+        customerId: String,
+        visitorId: String,
+        config: EmbeddedMessagingConfig,
+        message: EmbeddedMessage,
+        isRead: Bool
+    ) throws -> NetworkRequest {
+        
+        let (region, brandId, tenantId) = getConfigValues(from: config)
+        let baseURL = URL(string: "https://optimobile-inbox-srv-\(region).optimove.net")!
```
cgwyllie, line 395: Should we plug into the URL provider / is it done for pref center? (it allows overriding URLs to work locally etc., see something like set URL map in config IIRC)

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
daniel-c-opti: So I have added some helper methods as discussed that allow for custom types, by default it will just return a string and give the developer choice of how to deal with the JSON we get on that field: https://github.com/optimove-tech/Optimove-SDK-Android/pull/51/files#diff-e66d90b48bd0a7c993c926076b661f43e8d40af190e7bae4dfccc2db6ca5d6f6R109-R119

cgwyllie: Great. Did we change the shape of the get messages API to return the payload encoded? `payload = jsonMessage.optString("payload")`

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
cgwyllie: For public visibility, the purpose of the class is to expose a new feature (the data layer for embedded messaging), so I think it should remain public.

RE testing, yes, we'll need to pick this up.

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+    public let id: String
+    public let createdAt: Int
+    public let updatedAt: String?
+    public let deletedAt: String?
+}
+
+public typealias EmbeddedMessagesResponse = [String: EmbeddedMessagingContainer]
+
+
+public struct EmbeddedMessagingAPIResponse: Codable {
```
cgwyllie: Do all of these types need to be public or can some be internal?

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
daniel-c-opti: not sure I understand

cgwyllie: @daniel-c-opti was just meaning the payload given back by the API (I think for inbox) is just an object literal, not a string. Did we make EM read API JSON-encode the payload field into a string so the SDK reads it as `optString("payload")` in Java? (just clarifying my understanding)

daniel-c-opti: ah yeah it gets set as such in the constructor for the class where all of the objects get converted from JSON into their respective types.

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: we have stuck with this way for now to mirror preference centre, however could look to improve this in the furture

daniel-c-opti (approved):

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
eligutovsky: I agree about the data layer. 

But `EmbeddedMessagesService` should not be public, it is an internal service, right?

Shared/Sources/Mocks/MockOptimoveStorage.swift
```diff
@@ -8,6 +8,16 @@ public final class MockOptimoveStorage: OptimoveStorage {
...
+    
+    public var customerID: String? {
+        get { return value(for: .customerID) as? String }
+        set { set(value: newValue, key: .customerID) }
+    }
+
+    public var visitorID: String? {
+        get { return value(for: .visitorID) as? String }
+        set { set(value: newValue, key: .visitorID) }
+    }
```
eligutovsky: These two accessors are redundunt here

```swift
mockStorage.set(value: "adam_b@optimove.com", key: .customerID)
// equivalent to 
mockStorage[.customerID] = "adam_b@optimove.com"
```

Use [subscript technology](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/subscripts/) to avoid boilerplate code

eligutovsky: Check [the storage tests](https://github.com/optimove-tech/Optimove-SDK-iOS/blob/8075edc40726e3c00e832c6b250bf1a5dd570b41/OptimoveCore/Tests/Sources/Storage/OptimoveStorageFacadeTests.swift#L29) for usage examples.

eligutovsky: read values
```swift
let customerID = mockStorage[.customerID]
```

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
eligutovsky: > yeah there is likely a way to improve this, at this stage we are just following the same standards from preference center

I'm not sure what kind of standard it is. I believe we can do better, at least by following some code standards, e.g. Clean Code

Shared/Sources/Mocks/MockOptimoveStorage.swift
adam-b-optimove: Good shout, iv pushed this change

OptimoveSDK/Sources/Classes/EmbeddedMessaging/EmbeddedMessaging.swift
adam-b-optimove: updated this to be internal rather than public

cgwyllie: > it is an internal service, right?

ah yep, my bad, had skimmed the file and missed the two classes, derp

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
cgwyllie: @adam-b-optimove I think we need to address the type of the field here, for EM in future payload may not always be string,string.

cgwyllie: can we look at doing something similar to Android methods that offer it back as a JSON object or a generic type perhaps?

OptimoveSDK/Sources/Classes/EmbeddedMessaging/Models/EmbeddedMessage.swift
cgwyllie: Hey @adam-b-optimove @daniel-c-opti what was the outcome RE alignment?

PR #123, opened 2025-05-02
AB#247447 Handle query params in dynamic image urls

daniela-opti:
### Description of Changes

We found that urls with query parameters weren't working for push. The images were not displaying on the end user's device. 

The issue was that when URLs contained query parameters (like image.jpg?width=300), the URL's pathExtension property would incorrectly include those parameters, leading to improper extension extraction. The improved implementation separates the logic more clearly, trying both URL(string:) and fileURLWithPath approaches to ensure the correct file extension is extracted regardless of whether query parameters are present in the URL.

### Breaking Changes

-   None

### Prepare:

### Bump versions in:

- [ ] `OptimoveCore.podspec`
- [ ] `OptimoveSDK.podspec`

- [ ] `README.md`
- [ ] `CHANGELOG.md`

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:

### DIFF
--- a/OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
+++ b/OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
@@ -135,18 +135,17 @@ public class OptimoveNotificationService {
         })
     }
 
-    private class func getPictureExtension(_ pictureUrl: String?) -> String? {
+    internal class func getPictureExtension(_ pictureUrl: String?) -> String? {
         guard let pictureUrl = pictureUrl else {
             return nil
         }
-        
+
         if let url = URL(string: pictureUrl) {
             let pictureExtension = url.pathExtension
             return pictureExtension.isEmpty ? nil : "." + pictureExtension
         }
         
-        let pictureExtension = URL(fileURLWithPath: pictureUrl).pathExtension
-        return pictureExtension.isEmpty ? nil : "." + pictureExtension
+        return nil;
     }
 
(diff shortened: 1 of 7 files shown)


OptimoveNotificationServiceExtension/Sources/OptimoveNotificationService.swift
```diff
@@ -136,15 +136,17 @@ public class OptimoveNotificationService {
...
-            return nil
+        
+        if let url = URL(string: pictureUrl) {
+            let pictureExtension = url.pathExtension
+            return pictureExtension.isEmpty ? nil : "." + pictureExtension
         }
-
-        return "." + pictureExtension
+        
+        let pictureExtension = URL(fileURLWithPath: pictureUrl).pathExtension
```
daniela-opti: wasn't sure if this is needed? Thoughts

 maybe just return nil?

vvoicehovics: getPictureExtension is used in 2 places. It seems fileURLWithPath is not needed in both, but worth confirming what is `response?.suggestedFilename`

k-antipochkin:
Looks good. Just need to bump the versions and update the changelog

k-antipochkin (approved):

PR #140, opened 2026-04-13
AB#280444 - Add GamifyWidget SDK module

dmytro-b-optimove:
## Summary

- Adds `GamifyWidgetSDK` — public entry point with `initialize(widgetUrl:)` and `open(from:userId:token:)` that presents a modal sheet
- Adds `GamifyWidgetViewController` — internal UIKit view controller hosting a `WKWebView` with JS bridge (READY → INIT → CLOSE handshake), loading spinner, and error state
- Adds `GamifyWidgetSDKTests` — unit tests for URL initialisation

iOS equivalent of Android SDK PR #91 ([AB#280444](https://mobius.visualstudio.com/7b12c5d6-c2cb-4957-9a09-46dfabf0bd18/_workitems/edit/280444)). No changes to `Package.swift` — new files sit inside the existing `OptimoveSDK` target source path.

## JS Bridge

Widget fires `READY` via `window.webkit.messageHandlers.receiveMessage.postMessage({type: "READY", ...})` → native responds with `INIT` (userId/token) via `window.postMessage`. Widget calls `window.webkit.messageHandlers.closeWidget.postMessage(...)` to close. Handler names match the canonical bridge contract in [`opti-ls/packages/widget-fe/src/lib/bridge/bridge.ts`](https://github.com/optimove-tech/opti-ls/blob/dev/packages/widget-fe/src/lib/bridge/bridge.ts). Token is redacted in logs.

## Test plan

- [ ] `cmd+U` on `OptimoveSDK` scheme — `GamifyWidgetSDKTests` pass
- [ ] `cmd+B` on `iOS` scheme — clean build, no errors or warnings
- [ ] Run QA app, navigate to Gamify Widget, fill tenant + userId, tap Open Widget — sheet opens and close button dismisses

### DIFF
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -1,0 +1,48 @@
+// Copyright © 2024 Optimove. All rights reserved.
+
+import UIKit
+
+/// Entry point for the Gamify Widget SDK.
+///
+/// Usage:
+///   GamifyWidgetSDK.initialize(widgetUrl: "https://your-widget.example.com")
+///   GamifyWidgetSDK.open(from: viewController, userId: "u123")
+public final class GamifyWidgetSDK {
+
+    internal static var widgetUrl: String = ""
+
+    private init() {}
+
+    /// Configure the widget URL before opening.
+    public static func initialize(widgetUrl: String) {
+        self.widgetUrl = widgetUrl
+    }
+
+    /// Present the widget in a modal sheet.
+    ///
+    /// - Parameters:
+    ///   - viewController: The presenting UIViewController.
+    ///   - userId: Optional user ID injected via INIT handshake.
+    ///   - token: Optional auth token injected via INIT handshake.
+    public static func open(
+        from viewController: UIViewController,
+        userId: String? = nil,
+        token: String? = nil
+    ) {
+        let vc = GamifyWidgetViewController(
+            widgetUrl: widgetUrl,
+            userId: userId,
+            token: token
+        )
+        let nav = UINavigationController(rootViewController: vc)
+        if #available(iOS 15.0, *) {
+            if let sheet = nav.sheetPresentationController {
+                sheet.detents = [.large()]
+                sheet.prefersGrabberVisible = true
+            }
+        } else {
+            nav.modalPresentationStyle = .pageSheet
+        }
+        viewController.present(nav, animated: true)
+    }
+}
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -1,4 +1,4 @@
-// Copyright © 2024 Optimove. All rights reserved.
+//  Copyright © 2026 Optimove. All rights reserved.
 
 import UIKit
 
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -34,15 +34,14 @@ public final class GamifyWidgetSDK {
             userId: userId,
             token: token
         )
-        let nav = UINavigationController(rootViewController: vc)
         if #available(iOS 15.0, *) {
-            if let sheet = nav.sheetPresentationController {
+            if let sheet = vc.sheetPresentationController {
                 sheet.detents = [.large()]
                 sheet.prefersGrabberVisible = true
             }
         } else {
-            nav.modalPresentationStyle = .pageSheet
+            vc.modalPresentationStyle = .pageSheet
         }
-        viewController.present(nav, animated: true)
+        viewController.present(vc, animated: true)
     }
 }
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -15,6 +15,7 @@ public final class GamifyWidgetSDK {
 
     /// Configure the widget URL before opening.
     public static func initialize(widgetUrl: String) {
+        assertOnMainThread()
         self.widgetUrl = widgetUrl
     }
 
@@ -29,6 +30,7 @@ public final class GamifyWidgetSDK {
         userId: String? = nil,
         token: String? = nil
     ) {
+        assertOnMainThread()
         let vc = GamifyWidgetViewController(
             widgetUrl: widgetUrl,
             userId: userId,
@@ -43,5 +45,9 @@ public final class GamifyWidgetSDK {
             vc.modalPresentationStyle = .pageSheet
         }
         viewController.present(vc, animated: true)
+    }
+
+    private static func assertOnMainThread(_ message: String = "Must be on main thread") {
+        assert(Thread.isMainThread, message)
     }
 }
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -43,6 +43,10 @@ public final class GamifyWidgetSDK {
         token: String? = nil
     ) {
         assertOnMainThread()
+        guard !widgetUrl.isEmpty, URL(string: widgetUrl) != nil else {
+            Logger.error("GamifyWidgetSDK.open called with an invalid widgetUrl.")
+            return
+        }
         let vc = GamifyWidgetViewController(
             widgetUrl: widgetUrl,
             userId: userId,
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
@@ -15,6 +15,10 @@ public final class GamifyWidgetSDK {
 
     /// Configure the widget URL before opening.
     public static func initialize(widgetUrl: String) {
+        ensureMain { initialize_onMain(widgetUrl: widgetUrl) }
+    }
+
+    private static func initialize_onMain(widgetUrl: String) {
         assertOnMainThread()
         self.widgetUrl = widgetUrl
     }
@@ -26,6 +30,14 @@ public final class GamifyWidgetSDK {
     ///   - userId: Optional user ID injected via INIT handshake.
     ///   - token: Optional auth token injected via INIT handshake.
     public static func open(
+        from viewController: UIViewController,
+        userId: String? = nil,
+        token: String? = nil
+    ) {
+        ensureMain { open_onMain(from: viewController, userId: userId, token: token) }
+    }
+
+    private static func open_onMain(
         from viewController: UIViewController,
         userId: String? = nil,
         token: String? = nil
@@ -45,6 +57,14 @@ public final class GamifyWidgetSDK {
             vc.modalPresentationStyle = .pageSheet
         }
         viewController.present(vc, animated: true)
+    }
+
+    private static func ensureMain(_ work: @escaping () -> Void) {
+        if Thread.isMainThread {
+            work()
+        } else {
+            DispatchQueue.main.async(execute: work)
+        }
     }
 
     private static func assertOnMainThread(_ message: String = "Must be on main thread") {
--- a/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
+++ b/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
@@ -1,0 +1,158 @@
+// Copyright © 2024 Optimove. All rights reserved.
+
+import UIKit
+import WebKit
+
+private enum BridgeMessage {
+    static let handlerName = "nativeBridge"
+}
+
+final class GamifyWidgetViewController: UIViewController {
+
+    private let widgetUrl: String
+    private let userId: String?
+    private let token: String?
+
+    private var webView: WKWebView!
+    private var activityIndicator: UIActivityIndicatorView!
+    private var errorLabel: UILabel!
+
+    init(widgetUrl: String, userId: String?, token: String?) {
+        self.widgetUrl = widgetUrl
+        self.userId = userId
+        self.token = token
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) { fatalError() }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view.backgroundColor = .systemBackground
+        setupNavigationBar()
+        setupWebView()
+        setupLoadingIndicator()
+        setupErrorLabel()
+        loadWidget()
+    }
+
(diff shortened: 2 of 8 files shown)


SofiaVainstein (dismissed):

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    private func showError() {
+        DispatchQueue.main.async {
+            self.activityIndicator.stopAnimating()
+            self.webView.isHidden = true
+            self.errorLabel.isHidden = false
+        }
+    }
+
+    @objc private func dismissSelf() {
```
graciecooper: Do we need `@objc` here?

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        webView = WKWebView(frame: .zero, configuration: config)
+        webView.navigationDelegate = self
+        webView.translatesAutoresizingMaskIntoConstraints = false
+
+        view.addSubview(webView)
+        NSLayoutConstraint.activate([
+            webView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            webView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            webView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            webView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
```
graciecooper:
```suggestion
            webView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
```

Best to match what we've done for `topAnchor` maybe ?

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
```diff
@@ -0,0 +1,47 @@
+// Copyright © 2024 Optimove. All rights reserved.
```
graciecooper: 2026?

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
```diff
@@ -0,0 +1,47 @@
...
+import UIKit
+
+/// Entry point for the Gamify Widget SDK.
+///
+/// Usage:
+///   GamifyWidgetSDK.initialize(widgetUrl: "https://your-widget.example.com")
+///   GamifyWidgetSDK.open(from: viewController, userId: "u123")
+public final class GamifyWidgetSDK {
+
+    internal static var widgetUrl: String = ""
```
graciecooper, line 12: See this static var is mutated in `initialize()` and read in `open()` without sync. Could this race?? Since this is UIKit, maybe we could mark the class @MainActor? Lmk!

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        }
+        setupWebView()
+        setupLoadingIndicator()
+        setupErrorLabel()
+        loadWidget()
+    }
+
+    private func setupWebView() {
+        let contentController = WKUserContentController()
+        contentController.add(self, name: BridgeMessage.receiveMessage)
```
graciecooper, line 43: `WKUserContentController.add(_:name:)` retains self strongly, creating a reference cycle... I know we've not QA'd this yet with the test app, but could this maybe cause memory leak?

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
dmytro-b-optimove: Good catch on the race. I looked at how the existing codebase handles main-thread invariants (e.g. [`InAppPresenter`](https://github.com/optimove-tech/Optimove-SDK-iOS/blob/a8974016c7cd9494b9827c0edf8204c2191eec74/OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift#L628-L638)) — it uses `assertOnMainThread()` / `ensureMain { }` helpers rather than Swift Concurrency, and `@MainActor` isn't used anywhere in the SDK. Given the package targets Swift 5.3 / iOS 10, I'd prefer to stay consistent with the existing pattern rather than introduce `@MainActor` in just this one entry point. Happy to add an `assertOnMainThread()` to `initialize()` and `open()` instead — wdyt?

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
dmytro-b-optimove: Confirmed there is a retain cycle here, but I checked [`InAppPresenter.swift`](https://github.com/optimove-tech/Optimove-SDK-iOS/blob/a8974016c7cd9494b9827c0edf8204c2191eec74/OptimoveSDK/Sources/Classes/Optimobile/InApp/InAppPresenter.swift#L268) (the other WebKit-using class in this SDK) and it does the same thing — `contentController.add(self, name: ...)` directly, no weak proxy. The cycle is broken when the VC is dismissed and `webView` goes away. I went with the existing pattern for consistency, but happy to add a `WeakScriptMessageHandler` proxy here if that's the preferred fix — lmk!

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
dmytro-b-optimove: changed to 2026

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
dmytro-b-optimove: updated to `webView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)`

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
dmytro-b-optimove: removed `@objc`

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
graciecooper: Cool, yeah let's stick with the existing pattern !! Let's go with `assertOnMainThread()` instead 🥇

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
graciecooper: Ah cool, well we can test in the QA app and if needed we can add the handler proxy? But if it matches existing patterns and that hasn't caused an issue elsewhere then happy to leave it as it :)

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
dmytro-b-optimove: Added `assertOnMainThread()` to `initialize()` and `open()` — matches the existing helper shape in `InAppPresenter`.

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
```diff
@@ -0,0 +1,53 @@
...
+                sheet.prefersGrabberVisible = true
+            }
+        } else {
+            vc.modalPresentationStyle = .pageSheet
+        }
+        viewController.present(vc, animated: true)
+    }
+
+    private static func assertOnMainThread(_ message: String = "Must be on main thread") {
+        assert(Thread.isMainThread, message)
```
eligutovsky, line 75: In release builds (-O), the call is stripped out entirely and does nothing. 

If you want the same check to fire in release builds too, use precondition(Thread.isMainThread, message) instead — that one is always evaluated.​​​​​​​​​​​​​​​​

What is you intent here? To execute code on the main thread?
If yes, then suggesting to use this pattern instead of assertion:
```swift
if Thread.isMainThread {
    doWork()
} else {
    DispatchQueue.main.async { doWork() }
}

```

dmytro-b-optimove: Looking at how the rest of the SDK handles this in `InAppPresenter.swift` (lines 628-638), there's already an `ensureMain { ... } + assertOnMainThread()` pattern: public entry points auto-dispatch via `ensureMain` (e.g. lines 105, 169, 215), and the private `_onMain` body keeps `assertOnMainThread()` as a debug guard (e.g. lines 109, 173, 219).

Adopted that same pattern here in 13783a46 — wraps the bodies of `initialize(...)` and `open(...)` in `ensureMain`, which matches your suggested auto-dispatch snippet and stays consistent with the rest of the codebase. Kept `assert` (not `precondition`) since that's what `InAppPresenter` uses; once auto-dispatch is in place release-build enforcement isn't needed because the SDK guarantees main-thread itself.

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+        contentController.add(self, name: BridgeMessage.closeWidget)
+
+        let config = WKWebViewConfiguration()
+        config.userContentController = contentController
+
```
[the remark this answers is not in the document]
dmytro-b-optimove: following existing codebase approach here

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
+//  Copyright © 2026 Optimove. All rights reserved.
+
+import UIKit
+import WebKit
```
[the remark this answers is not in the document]
dmytro-b-optimove: Logger is already usable without an explicit import OptimoveCore
https://github.com/optimove-tech/Optimove-SDK-iOS/blob/feature/280444-gamify-widget-ios/OptimoveSDK/Sources/Classes/OptimoveConfig.swift

OptimoveSDK/Tests/Sources/GamifyWidget/GamifyWidgetSDKTests.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    func testInitializeSetsWidgetUrl() {
+        let url = "https://gamify-widget.example.com"
+        GamifyWidgetSDK.initialize(widgetUrl: url)
+        XCTAssertEqual(GamifyWidgetSDK.widgetUrl, url)
+    }
+
+    func testInitializeOverwritesPreviousUrl() {
+        GamifyWidgetSDK.initialize(widgetUrl: "https://first.example.com")
+        GamifyWidgetSDK.initialize(widgetUrl: "https://second.example.com")
```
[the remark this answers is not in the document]
dmytro-b-optimove: ensureMain is synchronous on main, and XCTest runs on main

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
```diff
@@ -0,0 +1,73 @@
...
+        from viewController: UIViewController,
+        userId: String? = nil,
+        token: String? = nil
+    ) {
+        assertOnMainThread()
+        let vc = GamifyWidgetViewController(
+            widgetUrl: widgetUrl,
+            userId: userId,
+            token: token
+        )
```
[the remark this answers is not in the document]
dmytro-b-optimove: Empty/invalid `widgetUrl` doesn't produce a blank sheet — `GamifyWidgetViewController.loadWidget()` guards via `URL(string:)` and falls through to `showError()`, which reveals the error label set up in `setupErrorLabel()` ([L80-L102](https://github.com/optimove-tech/Optimove-SDK-iOS/blob/feature/280444-gamify-widget-ios/OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift#L80-L102)):

> "Unable to load widget.\nCheck your connection and try again."

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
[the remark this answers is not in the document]
eligutovsky: Hmm, interesting how it was linked then

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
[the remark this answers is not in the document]
eligutovsky: How does this error message help a user? 
Copilot says not to open the error screen if we know that the user will see it immediately.

OptimoveSDK/Tests/Sources/GamifyWidget/GamifyWidgetSDKTests.swift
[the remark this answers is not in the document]
eligutovsky: The `ensureMain` is not synchronous, and it should stay async (to prevent blocking the UI thread). So there is a chance of a false-negative test run

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetViewController.swift
```diff
@@ -0,0 +1,160 @@
...
+
+private enum BridgeMessage {
+    static let receiveMessage = "receiveMessage"
+    static let closeWidget = "closeWidget"
+}
```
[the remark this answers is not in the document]
dmytro-b-optimove: iOS handler names match the canonical JS bridge in [`optimove-tech/opti-ls` → `packages/widget-fe/src/lib/bridge/bridge.ts`](https://github.com/optimove-tech/opti-ls/blob/dev/packages/widget-fe/src/lib/bridge/bridge.ts) — widget posts to `webkit.messageHandlers.receiveMessage` for general events and `webkit.messageHandlers.closeWidget` for close. `nativeBridge` doesn't exist anywhere in `opti-ls`; the PR description is stale. Will update the description; no code change.

OptimoveSDK/Sources/Classes/GamifyWidget/GamifyWidgetSDK.swift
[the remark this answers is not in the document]
dmytro-b-optimove: Updated it to match Copilot suggestion "fails fast and doesn't show a blank sheet to end users."

OptimoveSDK/Tests/Sources/GamifyWidget/GamifyWidgetSDKTests.swift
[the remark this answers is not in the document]
dmytro-b-optimove: Applied Copilot's suggested runOnMainSync helper so the two on-main tests force the initialize call onto main before asserting. Removes the off-main false-negative risk without changing ensureMain.
@eligutovsky check please

kris-opti (approved):

PR #141, opened 2026-06-01
AB#284943 - Action handler for Overlay Messaging

vvoicehovics:
### Description of Changes

Adds an **Overlay Messaging action handler** API, aligned with Android and Web (ADR-0001), so integrators can intercept overlay CTA clicks and perform their own deep-linking instead of the SDK opening the URL.

**Default behaviour (unchanged)**

When no handler overrides are registered, link buttons open via `UIApplication.shared.open` — universal/app links are handled by the OS as today.

**New API**

Conform to `OverlayActionHandlers` and pass an instance to `setActionHandlers`. Override only the actions you want to own; unoverridden actions keep the SDK default automatically.

```swift
class MyHandlers: OverlayActionHandlers {
    func linkAction(message: OverlayMessagingMessage, data: LinkActionData) {
        // Custom in-app routing — data.url is the resolved URL string
        // message.data is available for campaign context (e.g. tags)
    }
}

// At app startup
Optimove.shared.overlayMessaging.setActionHandlers(MyHandlers())

// Pass nil to clear overrides and restore all SDK defaults
Optimove.shared.overlayMessaging.setActionHandlers(nil)
```

**Handler contract**

- `Optimove.shared.overlayMessaging.setActionHandlers(_ handlers: OverlayActionHandlers?)` — pass a conforming type to override one or more actions; `nil` restores all SDK defaults
- `OverlayActionHandlers` protocol — one method per action type, each with a strongly-typed payload; implementing only the methods you need is sufficient
- `LinkActionData` — typed payload for link CTAs: `url: String`
- The SDK dispatches `overrides ?? defaults`, so any method not implemented on the client type runs the SDK default automatically
- If the handler throws, the error is logged and the SDK does **not** fall back to the default

**Internal changes**

- New `OverlayActionDefault.swift` — `DefaultOverlayActionHandlers` holds the SDK's built-in `linkAction` implementation (opens URL via `UIApplication.shared.open`)
- `OverlayActionDispatcher` — refactored from a per-type closure map to a defaults+overrides pattern; `setOverrides(_:)` stores `handlers ?? defaults`; `dispatch(_:message:)` always dispatches with no bool return
- `OverlayMessagingPresenter` — `handleSdkActions` passes a typed `OverlayAction.linkAction(LinkActionData)` to the dispatcher; the UIApplication fallback is removed (default handler owns that)
- `OverlayMessagingManager` — initialises dispatcher with `DefaultOverlayActionHandlers()`; exposes `setActionHandlers`

**Version:** 6.8.0

image

### Breaking Changes

- None (new API)

### Prepare:

- [x] Check `pod lib lint` passes

### Bump versions in:

- [x] `OptimoveCore.podspec`
- [x] `OptimoveSDK.podspec`
- [x] `README.md`
- [x] `CHANGELOG.md`
- [x] Update major version numbers in wiki (basic integration + push guides)

### Test Rail - Test run
https://optimove.testrail.io/index.php?/runs/view/850&group_by=cases:section_id&group_order=asc&display=tree

### Integration tests

*T&T Only*

- [ ] Associate customer
- [ ] Associate email
- [ ] Track events

*Mobile Only*

- [ ] Track events
- [ ] Register for push
- [ ] Opt-in for In-App
- [ ] Send test push
- [ ] Send test In-App
- [ ] **Overlay Messaging: link button opens URL when no `setActionHandlers` override is set**
- [ ] **Overlay Messaging: `setActionHandlers(MyHandlers())` — `linkAction` receives `OverlayMessagingMessage` and `LinkActionData(url:)`; SDK does not open the URL**
- [ ] **Overlay Messaging: `setActionHandlers(nil)` — default URL open restored**
- [ ] **Overlay Messaging: throwing `linkAction` handler — error is logged, URL is not opened**

*Deferred Deep Links*

*Combined*

### Release:

### Post Release:


OptimoveSDK/Sources/Classes/Optimobile/OverlayMessaging/OverlayMessagingPresenter.swift
```diff
@@ -124,19 +127,23 @@ final class OverlayMessagingPresenter: NSObject, WKScriptMessageHandler, WKNavig
     }
     
     private func destroyViews() {
+        contentController?.removeScriptMessageHandler(forName: "inAppHost")
+        contentController = nil
```
vvoicehovics, line 133: 2 minor fixes for overlay here as well:

1. presenter (added as self here `contentController!.add(self, name: "inAppHost")`) was leaking when views destroyed
2. remove navigation delegate during view destruction to prevent any potential races (e.g. error delegates running and decreasing slot count 2nd time). This is unlikely + slot count can only be 1 now, still safer.

adam-b-optimove (dismissed):

OptimoveSDK/Sources/Classes/Optimobile/OverlayMessaging/OptimoveOverlayMessaging.swift
```diff
@@ -72,3 +92,12 @@ public protocol OverlayMessagingInterceptor: AnyObject {
 public extension OverlayMessagingInterceptor {
     func getTimeoutMs() -> Int { 5000 }
 }
+
+public struct OptimoveOverlayMessagingAPI {
```
cgwyllie: What's the rationale for this API facade?

From what I can see the OptimoveOverlayMessaging class is already the 'public' entrypoint, does this add confusion/split in the API?

(correct me if I missed something)

stuart-reid-commits (approved):

